PR #53, opened 2018-01-11
feature/hash-cache-keys

### LINKED ISSUE #48: Optimize Cache Keys
Currently cache keys can be extraordinarily long, which might cause an issue if max key length is reached.

To prevent this, I had the idea of generating cache keys the same way they are now, but then using their hash as the key.

This will reduce key size, as well as obfuscate its content. (That would only be important for encrypted data, which it currently is not, but could also be a future consideration.)

I could see this being a quick little project for you. Please let me know your thoughts on the concept (pros/cons), and if you would be interested in it.

sjbronner:
fixes #48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor
 composer.lock
+.*.sw?
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 /vendor
 composer.lock
-.*.sw?
--- a/src/CachedBuilder.php
+++ b/src/CachedBuilder.php
@@ -17,9 +17,12 @@ class CachedBuilder extends EloquentBuilder
         }
 
         return $this->cache($this->makeCacheTags())
-            ->rememberForever($this->makeCacheKey() . "-avg_{$column}", function () use ($column) {
-                return parent::avg($column);
-            });
+            ->rememberForever(
+                $this->makeCacheKey(['*'], null, "-avg_{$column}"),
+                function () use ($column) {
+                    return parent::avg($column);
+                }
+            );
     }
 
     public function count($columns = ['*'])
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 /vendor
 composer.lock
+.*.sw?
```
mikebronner: This is probably best handled in your global git configuration. It's been considered "best practice" to leave out environment-specific ignores from the project's git-ignore.

src/CachedBuilder.php
```diff
@@ -123,11 +147,8 @@ public function pluck($column, $key = null)
             return parent::pluck($column, $key);
         }
 
-        $cacheKey = $this->makeCacheKey([$column]) . "-pluck_{$column}";
-
-        if ($key) {
-            $cacheKey .= "_{$key}";
-        }
+        $keyDifferentiator = "-pluck_{$column}" . ($key ? "_{$key}" : "");
```
mikebronner: More concise:
`$keyDifferentiator = "-pluck_{$column}_" . ($key ?: "");`

src/CachedModel.php
```diff
@@ -55,7 +55,7 @@ public static function all($columns = ['*'])
         $class = get_called_class();
         $instance = new $class;
         $tags = [str_slug(get_called_class())];
-        $key = 'genealabslaravelmodelcachingtestsfixturesauthor';
+        $key = sha1('genealabslaravelmodelcachingtestsfixturesauthor');
```
mikebronner: Call `makeCacheKey()` here, this was bad code on my part.

sjbronner: This won't work in a static function.

sjbronner: I'm confused as to why the function is marked static in the first place. It is called like an object method.

mikebronner: Something like this?
```php
$key = $instance->makeCacheKey();
```

sjbronner: for instance in `tests/Unit/CachedBuilderTest.php` line 44.

PR #56, opened 2018-01-12
Feature/fix phpunit colors

sjbronner:
The package `phpunit-result-printer` was not respecting `PHPUnit`'s specification of whether to colorize the output or not. This was a nuisance when piping the output of the unit tests to a pager or editor.
The updated version fixes that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php": ">=7.0.0"
     },
     "require-dev": {
-        "codedungeon/phpunit-result-printer": "^0.4.4",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -14,7 +14,7 @@
         "php": ">=7.0.0"
     },
     "require-dev": {
-        "codedungeon/phpunit-result-printer": "^0.4.4",
+        "codedungeon/phpunit-result-printer": "^0.5.3",
```
mikebronner: Do this instead: 
`"codedungeon/phpunit-result-printer": "*",`

PR #65, opened 2018-02-11
Move the logic into a trait, so it is accessible without extending the provided base model (eg for projects which already have a base model)

rs-sliske:
mostly just moving the code into the cachable trait, altohugh also cleaned up the boot method a little

the model instance gets passed to the event listeners by laravel anyway, and saved gets called for all the events so no need to handle created/updated/deleted as well


mikebronner:
Thanks @rs-sliske ! I've been thinking about this as well, you beat me to it! :) I'm traveling until Tuesday, but will take a look at it as soon as I get back to the office.

PR #74, opened 2018-02-16
Missing WhereNotIn method of Eloquent model

smskin:
Need add "NotIn" element to array in getTypeClause method


mikebronner:
Please also provide a unit test that addresses this specific issue. If you're not sure how, I can add it before merging in your PR. Any issues should be replicated by a failing test, before being fixed, so that the code is protected from that issue going forward. :)

PR #80, opened 2018-02-18
remove unneeded check to make compatible with mongodb

vesper8:
In order for my mongodb models to be able to extend this trait, I had to make a CachedModel of my own as seen below

```
use GeneaLabs\LaravelModelCaching\CachedBuilder as Builder;
use GeneaLabs\LaravelModelCaching\Traits\Cachable;
use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
use Illuminate\Database\Eloquent\Model;
use Jenssegers\Mongodb\Eloquent\Model as Eloquent;

abstract class _CachedMongoModel extends Eloquent
{
    use Cachable;

    ... rest is the same as your CachedModel
}
```

But then the above check was a blocker. Upon inspection it seems like there's no reason to make this check since we can assume anyone using your trait is using it with your CachedModel (or an adapted clone as in my case)

### DIFF
--- a/src/Traits/Cachable.php
+++ b/src/Traits/Cachable.php
@@ -21,9 +21,7 @@ trait Cachable
         }
 
         if (is_subclass_of($cache->getStore(), TaggableStore::class)) {
-            if (is_a($this, CachedModel::class)) {
-                array_push($tags, str_slug(get_called_class()));
-            }
+            array_push($tags, str_slug(get_called_class()));
 
             $cache = $cache->tags($tags);
         }


mikebronner:
@vesper8 Thanks for submitting this PR! Unfortunately I don't have a Laravel project using MongoDB to test this with. I won't be able to merge it unless all tests are passing.

vesper8:
hrm ok.. perhaps that check is not as unneeded as I thought. Even though commenting it out in my fork does not seem to be causing any issues.

Perhaps the solution then is to add a configuration option where one can whitelist which classes are recognized, with the default being the included CachedModel class

I guess for now I will just re-fork with your latest changes and hardcode it like this

```
            if (is_a($this, CachedModel::class) || is_a($this, CachedMongoModel::class)) {
                array_push($tags, str_slug(get_called_class()));
            }

```

mikebronner:
@vesper8 I was looking at your change and realized you were right, but I need to analyze the ramifications first (that's why I deleted my comment). I will take a look at that now.

mikebronner:
If you wish to join, I'm live-streaming this: https://www.larastream.com/r/40555b5 :)

vesper8:
I joined, but not sure if you can see my comments

PR #81, opened 2018-02-18
Add multitenant support to allow keeping tenant specific data separated for the cached models.

lucian-dragomir:
This is my proposal of adding multi-tenant support to the cached models by implementing at model level (in a base model class or using a trait) a new method getCachePrefix that can be used to differentiate cache tags and keys by prefixing them with the tenant context (in the provided example the method is returning the connection name concatenated with the tenant database name).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ relationships. This package is an attempt to address those requirements.
 -   automatic, self-invalidating model query caching.
 -   automatic use of cache tags for cache providers that support them (will
     flush entire cache for providers that don't).
+-   support for multitenant implementations by implementing getCachePrefix method in the Model class 
 
 ## Requirements
 -   PHP >= 7.1.3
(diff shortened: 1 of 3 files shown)


mikebronner:
I would recommend creating individual redis connections for each tenant, and handle it via configuration, instead of  implementing it in the code.

mikebronner:
Sorry, didn't mean to close it.

mikebronner:
I've been reading through you proposed changes some more, and feel this PR contains too many different changes (this that are not necessarily related to just implementing cache prefixing, but other refactoring as well). Would you mind stripping this PR down to only cache prefixing without any of the other refactoring, if possible? I don't necessarily want to merge all those changes.
If that's a problem, no worry, I'll implement cache prefixing soon.

mikebronner:
Actually, sorry to flip-flop on this so much. I will be implementing cache key prefixes now, but not using this PR, as I don't want to incorporate everything it has. Thanks for submitting this, though. Sorry for not accepting your PR, hopefully next time! :)

mikebronner:
Hi @lucian-dragomir this should be fixed in version 0.3.31. Please let me know how that works for you.

lucian-dragomir:
Hi @mikebronner, 
I read your recommendation to use different cache connection for each tenant and I think it is the correct and safe approach.
Thank you also for the cache prefix feature!

PR #89, opened 2018-02-24
typo fix in config/laravel-model-caching.php

rssawhney:
currently not retrieving the correct variable from .env file


mikebronner:
!!! Thanks for catching that!! sheesh :)

PR #97, opened 2018-03-08
dont cache if model::$isCachable == false

hxnk:
When calling model::all(), the $isCachable property on the model wasn't checked, resulting in model:all() results to be cached, even when $isCachable was set to false. 

This PR attempts to fix that :)

### DIFF
--- a/src/Traits/ModelCaching.php
+++ b/src/Traits/ModelCaching.php
@@ -7,12 +7,13 @@ trait ModelCaching
 {
     public static function all($columns = ['*'])
     {
-        if (config('laravel-model-caching.disabled')) {
+        $class = get_called_class();
+        $instance = new $class;
+
+        if (config('laravel-model-caching.disabled') || !$instance->isCachable) {
             return parent::all($columns);
         }
 
-        $class = get_called_class();
-        $instance = new $class;
         $tags = [str_slug(get_called_class())];
         $key = $instance->makeCacheKey();
 
--- a/src/Traits/ModelCaching.php
+++ b/src/Traits/ModelCaching.php
@@ -1,6 +1,7 @@
 <?php namespace GeneaLabs\LaravelModelCaching\Traits;
 
 use GeneaLabs\LaravelModelCaching\CachedBuilder;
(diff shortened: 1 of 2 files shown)


src/Traits/ModelCaching.php
```diff
@@ -7,12 +7,13 @@
 {
     public static function all($columns = ['*'])
     {
-        if (config('laravel-model-caching.disabled')) {
+        $class = get_called_class();
+        $instance = new $class;
+
+        if (config('laravel-model-caching.disabled') || !$instance->isCachable) {
```
mikebronner: The problem is that `isCachable` can never be set for the `all()` method, since setting `isCachable` is set through a query scope, which returns a Builder object, which doesn't have an all method. Consider that the following code throws an exception:
```php
$model->disableCache()->all();
```

Unfortunately the only way to disable caching on `all()` queries is to turn it off in config at this point, or change to `get()`. Perhaps you have some thoughts on how we can disable the cache on the model? In the past, I've had a method on the model, instead of a scope, and that had more problems than it solved, so I moved to scope.

mikebronner:
Hi @hxnk, thank you for submitting this PR! :) I left a comment explaining why this code won't work at this point. Please let me know if you have additional ideas. :)

hxnk:
Hi @mikebronner, what do you think of this solution? It's a bit hacky, but it works :)

- disableCache() temporarily disables cache in the config and returns a QueryOrModelCaller object, that is able to direct chained calls to the right objects (all() to Builder, rest to Model). This allows using a query scope and still chain the all() call, which would normally not be possible since a Builder object would be returned
- in the QueryOrModelCaller object cache setting in the config is set back to its original state, so that subsequent queries aren't affected

mikebronner:
Hi @hxnk, I initially thought that might work as well, but the realized that any global method of disabling cache would cause unexpected behavior for other possible simultaneous request. To avoid uninterrupted side-effects, we will have to find a solution that works on the model instance.

mikebronner:
Hi @hxnk  -- I was able to come up with a solution! :) Thanks for keeping the discussion going around this challenge and prompting me to think about different ways of achieving this!

PR #120, opened 2018-04-28
Multiple Database Connections: Extension

bastian-schur:
I changed the connection name to the database name. This would be more flexible if a tenant system use for every tenant the same connection name and only changed the database in the connection.
This change has no disadvantages compared to the previous variant


bastian-schur:
I'm not sure if the problems result in this change?

mikebronner:
Yes, the failures are directly related to this PR. I do have plans to add the database name to the cache key, but keeping the connection name is critical as well.

PR #121, opened 2018-05-02
formatting

shenjiayu:


mikebronner:
Thanks for submitting a PR, but formatting changes are generally not useful, especially when the format of the code is intentional. I would welcome Functional PRs that address some of the current issues. :)

Issue #139, opened 2018-06-05
Undefined offset in conjunction with yajra/laravel-datatables


rappasoft:
# Issue
This issue prevents searching on the laravel datatables plugin due to some type of issue with the laravel datatables usage of whereRaw.

Possibly related to: https://github.com/GeneaLabs/laravel-model-caching/issues/111

Possible clue:

```
DataTables::of(User::query()) -> Error
DataTables::of(\DB::table('users')) -> Works
```

## Environment
Laravel Version: 5.6.24
Laravel Model Caching Package Version: 0.2.62
PHP Version: 7.2
Homestead Version: 6
Operating System & Version: MacOS 10.13.5

## Stack Trace
```
[2018-06-03 19:34:22] local.ERROR: ErrorException: Undefined offset: 3 in /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php:235
Stack trace:
#0 /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php(235): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Undefined offse...', '/home/vagrant/C...', 235, Array)
#1 /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php(152): GeneaLabs\LaravelModelCaching\CacheKey->getRawClauses(Array)
#2 [internal function]: GeneaLabs\LaravelModelCaching\CacheKey->GeneaLabs\LaravelModelCaching\{closure}('-_or_LOWER(`pra...', Array)
#3 /PROJECT_PATH/vendor/laravel/framework/src/Illuminate/Support/Collection.php(1313): array_reduce(Array, Object(Closure), NULL)
#4 /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php(158): Illuminate\Support\Collection->reduce(Object(Closure))
#5 /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php(169): GeneaLabs\LaravelModelCaching\CacheKey->getWhereClauses(Array)
#6 /PROJECT_PATH/vendor/genealabs/laravel-model-caching/src/CacheKey.php(150): GeneaLabs\LaravelModelCaching\CacheKey->getNestedClauses(Array)
[...]

PR #147, opened 2018-07-09
Fix nested where binding issue

ikerasLT:
Removed unnecessary binding advancement. 
Solves issue with yajra/laravel-datatables global search (#139)


mikebronner:
Thanks for posting this. I will review this over the weekend hopefully, as soon as I can. :)

PR #180, opened 2018-10-28
Change modelCache command to flush

d8vjork:
As described in the title, as I see now is `modelCache:flush` instead of `modelCache:clear`


mikebronner:
Hi @d8vjork, thanks for submitting.
Using the `clear` command is intentional, so as to match the other `clear` commands that Laravel has.

PR #192, opened 2018-11-23
Resolving issue when model has dynamic table, table name added i…

jetmirhaxhisefa:
…n getModelSlug

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -59,7 +59,7 @@ class CacheKey
 
     protected function getModelSlug() : string
     {
-        return str_slug(get_class($this->model));
+    	return str_slug(get_class($this->model).":".$this->model->getTable());
     }
 
     protected function getOffsetClause() : string


mikebronner:
@jetmirhaxhisefa thanks for submitting this. Can you please prove a code example that requires this change? We need to build out the integration tests to make sure this code works as intended in future changes as well. Thanks!

jetmirhaxhisefa:
@mikebronner here is my problem. 
I have a model "Post" but multiple post tables for example "post_articles", "post_pages", "post_testimonials" etc.
I use the same Post model to query dynamically in these tables:

      $posts  = (new Posts())->setTable($table)->get();

The problem is when the CacheKey is generated it doesn't have a table identifier. So if for example, I used:

      $articles  = (new Posts())->setTable("post_articles")->get();

cache get stored for the post model with get() and than I try to use same model with a different table like: 

      $pages  = (new Posts())->setTable("post_pages")->get();

$pages will return **articles** items because it is stored previously in cache and it doesn't recognize that it uses a different table.

mikebronner:
Thanks for the feedback. I will close this PR, as I want to implement it differently, but not to worry, I will add it. Thanks!

Issue #212, opened 2019-02-19
Multitenant cache - cache fetched for wrong tenant


denitsa-cm:
**Describe the bug**
Tenant connection fetches the cache for a differnet tenant.

**Eloquent Query**

The query I'm executing for a `/settings` request:

```php
return collect(Member::first()->only('name', 'alternative_name', 'abbreviation', 'ringgold_id', 'repository', 'website', 'vat'))
            ->merge(MemberSettingKey::all()->pluck('value', 'slug'));
```

**Environment**
 - PHP: 7.2
 - OS: Ubuntu 18.04
 - Laravel: 5.7.24
 - Model Caching: 0.3.4

**Additional context**
We have a `main` connection and a `tenant` connection. The specific tenant database is set in a middleware on every request. 

We just had an issue where fetching the settings of one tenant uses the cache of another.
I'm attaching a zip of the cache that is created after the request made by the first tenant.
Sending the same request for another tenant returns the data of the first.

[data.zip](https://github.com/GeneaLabs/laravel-model-caching/files/2879134/data.zip)

I can see that in the 7b->b4 folder there is a key that has the database name `tenant:.....` (which looks correct) 🤔 

My `MemberSettingKey` model extends `SettingKey` which in turn extends `BaseTenantModel`. The `Cachable` trait is applied to the `SettingKeyModel`. The `BaseTenantModel` only contains this:

```
class BaseTenantModel extends BaseModel
{
    /**
     * Create a new Eloquent model instance.
     *
     * @param  array $attributes
     * @return void
     */
    public function __construct(array $attributes = [])
    {
        $this->connection = config('database.tenant');
        parent::__construct($attributes);
    }
}
```

Would you be so kind to help me with this? :) 

P.S. I'm using `CACHE_DRIVER=file`

denitsa-cm:
Do I have to explicitly use the `CachePrefixing` trait somewhere? I see there's a function `CachePrefixing:getCachePrefix()` as well as `Caching::getCachePrefix()` 🤔

denitsa-cm:
I also dumped inside the `ModelCaching::all()` function. 

the result of  `$key = $instance->makeCacheKey();` is `genealabs:laravel-model-caching:main:librarymanager:appmodelsmembersettingkey` (which is wrong) 🤔

PokeGuys:
I am having the same issue here. 
[...]

PR #213, opened 2019-02-25
Fix incorrect cache key in multiple DB connection.

PokeGuys:
This PR fixed a `makeCacheKey` bug mentioned in Issue #212.
It used default DB connection to make query instead of using `$connection` in model.

### DIFF
--- a/src/Traits/CachePrefixing.php
+++ b/src/Traits/CachePrefixing.php
@@ -14,11 +14,11 @@ trait CachePrefixing
 
     protected function getDatabaseConnectionName() : string
     {
-        return $this->query->connection->getName();
+        return $this->model->getConnection()->getName();
     }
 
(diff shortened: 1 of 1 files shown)


mikebronner:
Thanks @PokeGuys ! I will take a look at this as soon as I can. :)

PokeGuys:
This will cause another issue. The eager load relation will be used the caller connection, not their own connection name.

mikebronner:
@PokeGuys thanks for the heads up, I'm close to wrapping up the L5.8 update, and will look at these next.

mikebronner:
@PokeGuys  Thanks for your patience, and bringing this issue up! This will be in the next release. I didn't merge, because there was a bunch of other work that needed to be done to allow model relationships to cache correctly across different connections and databases.

Issue #204, opened 2019-01-25
Bug: collection data and single data use same key


zhenyangze:
**Describe the bug**
laravel model return two type data: collection and  single object. if the laravel-model-caching use the same cache key, it will return one type of two. so when you want a collection data to display with foreach,it some times  report error because it return a single object. or you want to use a object property but it report the object doesn't have the property because it return collection data;

**Eloquent Query**
before use laravel-model-caching:
```php
// this will return a object data
$model->find(1) 

// this will return a collection data
$model->find([1])
```
after use laravel-model-caching:
```php
// this will return a object data,because it create the same key first:-find_1
$model->find(1) 

// this will return a object data,because it use the same key: -find_1
$model->find([1])
```
or 
```php
// this will return a collection data,because it create the same key first:-find_1
$model->find([1]) 

// this will return a collection data,because it use the same key: -find_1
$model->find(1)
```

**Stack Trace**
no error report ,bug get the data not expected.

**Environment**
 - PHP: [e.g. 7.1.0]
 - OS: [e.g. Ubuntu 18.04]
 - Laravel: [e.g. 5.6.15]
 - Model Caching: [e.g. 0.2.61]

**Additional context**
```php
    public function find($id, $columns = ["*"])
    {
        if (! $this->isCachable()) {
            return parent::find($id, $columns);
        }
        $idKey = collect($id)->implode('-');
        $cacheKey = $this->makeCacheKey($columns, null, "-find_{$idKey}");
        return $this->cachedValue(func_get_args(), $cacheKey);
    }
```

mikebronner:
I will have to look into this some more, before providing an accurate answer. Thanks for bringing this to my attention. :)

calchen:
@mikebronner I have the same problem.

mikebronner:
@calchen This should no longer be happening. Please test again with the latest version, 0.4.12 or above. Feel free to re-open if it does. Thanks!

PR #214, opened 2019-02-26
Laravel may return either an object or a collection. If laravel-model…

cluebattery:
…-caching uses the same cache key, it will return one of the two. See the comments by @zhenyangze at https://github.com/GeneaLabs/laravel-model-caching/issues/204#issue-402979412. This is just a simple commit to make a pull request for his recommended fix.


mikebronner:
Thanks @cluebattery, I will take a look at this as soon as possible. :)

Issue #209, opened 2019-02-07
Performance issue with `checkCooldownAndRemoveIfExpired`


yemenifree:
**Describe the bug**

I was testing the performance of the website with a huge number of requests in sametime but I found very critical issue with performance because of `checkCooldownAndRemoveIfExpired` function

its increase the response time from < 1s to > ~9s for each request when the server tried to serve > 400 requests per second, my test was done using different settings and using load balance for the web server and cache servers

after removing `checkCooldownAndRemoveIfExpired` from source code and use only model cache the server serves more than 2000 request per second without any issues or effect in response time

by the way, I don't use `cool-down` in my code

**Environment**

 - PHP: 7.2
 - OS: ubuntu 18.04.1
 - Laravel: 5.6
 - Model Caching: 0.3.6
 - 2 web servers
 - more than 500 request per second

## Perfomrnace with `checkCooldownAndRemoveIfExpired`

![image](https://user-images.githubusercontent.com/10876587/52403569-0a8fa780-2ad8-11e9-8556-b8a838176d86.png)

![image](https://user-images.githubusercontent.com/10876587/52404449-15e3d280-2ada-11e9-84c9-575b2e94689f.png)

## Perfomance without `checkCooldownAndRemoveIfExpired`

![image](https://user-images.githubusercontent.com/10876587/52403644-3dd23680-2ad8-11e9-98c1-de3bf51a4050.png)

mikebronner:
Thanks for submitting this in detail! I will take a further look into this, hopefully this weekend. :)

mycarrysun:
I can confirm - I found the same thing by analyzing the cache misses and that it is looking for the cooldown data even if it isn't there.

Are we able to add this to the config file somehow? 

I may write up a PR if I have some time.

mikebronner:
Yes, a PR would help out tremendously. I haven't had a chance to analyze this issue yet. :(

mycarrysun:
@mikebronner Just added - let me know what you think. The tests are failing somewhere before my tests run so not sure what's going wrong there.

Says: 
```
1) GeneaLabs\LaravelModelCaching\Tests\Integration\CachedBuilder\WhereJsonContainsTest::testWithInUsingCollectionQuery
Illuminate\Database\QueryException: SQLSTATE[08006] [7] timeout expired (SQL: select tablename from pg_catalog.pg_tables where schemaname = 'public')
```

mikebronner:
[...]

PR #223, opened 2019-03-24
Add feature to disable cooldown checking

mycarrysun:


mycarrysun:
This PR fixes #209 by adding an option to completely disable all cooldown checking, or by individual classes

mikebronner:
Thanks for working on this! I will review and try to get this (or whatever solution) integrated next week! I appreciate the help :)

mikebronner:
Hi @mycarrysun, I went for a slightly simpler (or at least I believe it is) solution, and because of that didn't integrate your PR. However, please don't let that be a sign that I don't appreciate your work. Thanks for submitting this, it helped me in reviewing the functionality and coming up with a solution I felt good with. Thanks! :)

mycarrysun:
Where are the changes that you implemented?

mikebronner:
See my comment on issue #209 ... check latest release:

https://github.com/GeneaLabs/laravel-model-caching/commit/f8aafe23a5a8b815b2fddac0a9f9595bb56d6136

mycarrysun:
Looks good to me - any reason you decided to go with a prop instead of a trait or in the config? Thinking people will be changing this at runtime?

mycarrysun:
Also good work on this library!!! I can't believe how much it improved the load time on our app....I knew it would be an improvement but didn't expect that much!!

Was thinking another feature that might be handy for folks using the cooldown, set a default cooldown time on the Model. Then you don't have to call it on every query.

mikebronner:
No, I wanted the setup to be declarative at the model level, since it's model-specific. I also did not want to introduce back-ward-incompatibility, I felt the config setting was an extra unnecessary step. Also, having empty traits on models to act as a flag seems odd to me, I would rather have a property be the flag, it is self-describing, you don't wonder what it does without having to go to a different file to look at it.

Using a default property will make it easy to implement for the few people that actually want to use it, and keep it completely out of the way of those that don't. :)

mikebronner:
Good call with the default cool down. In fact, I can combine that with the feature flag. If cool down is `0`, `null`, or `false`, then skip cooldown-checks, and if greater than `0`, then run them. I like it!

PR #225, opened 2019-03-27
Handle whereIn when collection is empty

Ben52:
Currently, when running a `whereIn`, where the collection checking against is empty an exception will be thrown. This fixes the issue, evidenced by the test.

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -215,7 +215,7 @@ class CacheKey
 
         $type = strtolower($where["type"]);
         $subquery = $this->getValuesFromWhere($where);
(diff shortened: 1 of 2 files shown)


mikebronner:
@Ben52  Thanks for submitting this! I will look into this soon. :)

mycarrysun:
@mikebronner this is related to #227 and PR #229 

I'm not sure, but I am thinking this PR will not work because it is still going to increase the `currentBinding` index. Either that or I'm missing something on how the `currentBinding` property works...

mikebronner:
@mycarrysun Looks like you were right. I've been getting bug reports with the bindings being off. Will need to re-investigate.

PR #228, opened 2019-03-28
Replace deprecated functions for Laravel 5.9

mycarrysun:


mycarrysun:
@mikebronner I don't think this is a facade, just a static method in the `Str` class

mikebronner:
@mycarrysun Haha, Even worse :) I never use statics unless I am forced to by the framework. (It's an opinionated stance, I know, but statics are a hold-over from procedural programming.)

mycarrysun:
Lol...not sure if you looked at the source code but it is using it whether you like it or not ;)

```php
/**
     * Generate a URL friendly "slug" from a given string.
     *
     * @param  string  $title
     * @param  string  $separator
     * @param  string  $language
     * @return string
     *
     * @deprecated Str::slug() should be used directly instead. Will be removed in Laravel 5.9.
     */
    function str_slug($title, $separator = '-', $language = 'en')
    {
        return Str::slug($title, $separator, $language);
    }
```

mikebronner:
@mycarrysun I understand the framework uses facades. My point is that code in my plugins should not use facades.

Issue #231, opened 2019-03-28
ModelCaching::scopeDisableCache() not working correctly


mycarrysun:
**Describe the bug**
Trying to use `disableCache()` on a query with eager loaded relationships only disables cache for the model, not the relationships

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
$order = Order::disableCache()->with([
    'company.bill_addr',
    'applied_payments.payment.card',
    'applied_payments.payment.applied_payments',
    'staff',
    'files',
    'events.user',
    // ...
])->findOrFail( $id );
```

**Stack Trace**
NA

**Environment**
 - PHP: 7.2
 - OS: Ubuntu 18.10
 - Laravel: 5.8.8
 - Model Caching: 0.4.11

**Additional context**
If I set the `MODEL_CACHE_DISABLED=true` the query runs correctly.

I also am getting hits on cache keys:
```
0: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appcompany",…],…}
1: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appappliedpayment",…],…}
2: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appfile"],…}
3: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appevent",…],…}
4: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appcarrier",…],…}
5: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:apprelatedorder",…],…}
6: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:apprelatedorder",…],…}
7: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:apppermit"],…}
8: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appcomment"],…}
9: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appcomment"],…}
10: {tags: ["genealabs:laravel-model-caching:mysql:ac09n332_dev:appcompany"],…}
```

mycarrysun:
[...]

Issue #106, opened 2018-03-28
Global scopes do not work with cachable models


scottgrayson:
# Issue
I would like to have a cache for users.index queries because they take a while.  I want to ->disableCache() on all other queries.  I tried a global scope that disables cache but found that global scopes do not work with models that use the `Cachable` trait.

This does not work:
- neither the `enabled` nor `nocache` scopes work here
- `enabled` works if I remove the `Cachable` trait
```php
    use Cachable;

    public static function boot()
    {
        parent::boot();

        static::addGlobalScope('enabled', function (CachedBuilder $builder) {
            // dd('hi');
            $builder->where('enabled', 1);
        });

        static::addGlobalScope('nocache', function (CachedBuilder $builder) {
            $builder->disableCache();
        });
    }
```

- I get no errors, but results always use cache.
- I never get the `dd('hi');` output if i uncomment it (shows that the callback given to addGlobalScope is never applied)

## Environment
Laravel Version: *5.6.10*
Laravel Model Caching Package Version: *0.2.52*
PHP Version: *7.1.11*
Valet Version: *2.0.7*
Operating System & Version: *MacOS 10.12.6*

## Stack Trace
N\A

mikebronner:
Hi @scottgrayson, thanks for reporting this use-case. I don't think this package is a good fit for you, if you only want to cache a single query in your entire app. You would be better off writing the cache functionality manually.

However, if you are using this on all other models, but only wish to have special treatment on a single model, my suggestion would be similar to above: do not make that model Cachable, but instead manually cache that one query as needed.

I will look into why the scope would not be working and see if there is a fix for that.

mikebronner:
PS: have you tried using a local scope on the model?

scottgrayson:
> However, if you are using this on all other models, but only wish to have special treatment on a single model, my suggestion would be similar to above: do not make that model Cachable, but instead manually cache that one query as needed.

This describes my situation. I'm working on caching it manually now, but if I could use your package for users as well, it would be a cleaner solution for me.  

> have you tried using a local scope on the model?

I have tried a local scope and it does work. 
```php
    public function scopeNocache($query)
    {
        return $query->disableCache();
    }
```
this is not very useful though. is there a different way you thought of using a local scope?

mikebronner:
[...]

PR #232, opened 2019-03-29
When model cache disabled, it is correctly disabled on query and all eager loaded relations

mycarrysun:
This is an attempt to fix the reason why `scopeDisableCache()` global scope is not working.

@mikebronner if this is supposed to be actually set to `true` for some reason, let me know. I believe it is meant to be false though.

See #106 and #231


mikebronner:
Oops, yea, looks like that might have been a blooper.

mycarrysun:
Closes #231

PR #288, opened 2019-09-09
🚧 Allow flushing by database name.

### LINKED ISSUE #287: Flush single database in multi-tenant application
When manually flushing the cache all cache is invalidated. It would be useful to only flush the current database connection or have the option to pass a database name.

mikebronner:
Fixes #287


mikebronner:
Currently not possible.

Issue #292, opened 2019-09-18
Invalidate Non-Cacheable Models


Hornet-Wing:
Firstly, thank you very much for this great package!

**Is your feature request related to a problem? Please describe.**
While using this package in my project I have found that some of my models which do not have the `Cacheable` trait, for example the `User` model, get cached as part of any eager loads with models that have the `Cacheable` trait.

For example:

Team: **has Cacheable**
User: **doesn't have Cacheable**
```php
App\Team::with('users')->find(1);
```

The above will cache the result which will contain the team details as well as all user details. The problem arises when a user changes their details. When the above is rerun the result will not reflect the changes as the cache was not invalidated.

I believe this is because the `User` model doesn't have the `Cacheable` trait in the first place, it will also not have the `ModelCaching` trait which adds the necessary listeners to invalidate the cache with the model's tag.

**Describe the solution you'd like**
I believe what is required is a generic trait that can be applied to all the models (which do not have the `Cacheable` trait) within the application so that all changes will cause a cache invalidation for that tag.

I had a quick go at how that trait might look just to demonstrate the principle. There are still a few kinks I still need to work out, such as just using the `Caching` trait doesn't quite work.

```php
<?php

namespace App\Traits;

use Illuminate\Support\Facades\Artisan;
use GeneaLabs\LaravelModelCaching\Traits\Caching;

trait InvalidateCacheable 
{

    use Caching;

    protected static function bootInvalidateCacheable() 
    {
        static::created(function ($instance) {
            $instance->clearModelCache();
        });

        static::deleted(function ($instance) {
            $instance->clearModelCache();
        });

        static::saved(function ($instance) {
            $instance->clearModelCache();
        });
    }

    protected function clearModelCache()
    {
        $this->flushCache();

        // OR

        // Artisan::call(
        // 	'modelCache:clear',
        // 	[
        // 		'--model' => get_class($this),
        // 	]
        // );
    }
}
```

**Describe alternatives you've considered**
An alternative would be to `disableCache` for all queries that reference a model that does not have the Cacheable trait. This is a perfectly feasible work around, however, it does rely on always remembering to do this which I don't believe is possible, especially in a team of developers.

I am happy to work on this if it makes sense and I haven't missed something that already exists in the package.

mikebronner:
@Hornet-Wing Thanks for writing this up with all the details. I will look into this over the next few weeks (I've been out of the office for the past 2 weeks).

[...]

PR #297, opened 2019-10-12
Testing for clearing of eager loaded non cacheable relations

Hornet-Wing:
This is a test to demostrate the bug discussed in #292. I have tried to use existing test classes and models but I can adapt it to create bespoke ones. Let me know if this is OK.

### DIFF
--- a/tests/Integration/CachedBuilderRelationshipsTest.php
+++ b/tests/Integration/CachedBuilderRelationshipsTest.php
@@ -44,4 +44,27 @@ class CachedBuilderRelationshipsTest extends IntegrationTestCase
 
         $this->assertEquals($books->pluck("id"), $uncachedBooks->pluck("id"));
     }
+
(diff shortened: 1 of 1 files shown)


mikebronner:
@Hornet-Wing Awesome! I will try to get to it this weekend. :)

mikebronner:
@Hornet-Wing Go ahead and give release 0.7.2 a try. The final implementation ended up being that any queries that have non-catchable eager-loaded relationships will not be cached in their entirety to avoid the cache of the relationship going stale. There really was no other way around that.

Hornet-Wing:
Sorry it took me a little while to get this tested. Unfortunetly my main project is still 5.8 so I created a seperate project to test this in isolation.

Unfortunetly it still doesnt quite work as expected. I'm planning on taking a closer look, maybe my tests were wrong. :/

https://github.com/Hornet-Wing/model-cache-eager-load

PR #302, opened 2019-10-23
Unification of whether to turn on cache detection

titrxw:


mikebronner:
Thanks for submitting this. I will take a look at this as soon as possible.

Issue #280, opened 2019-08-13
Issue With BelongsToMany


mikebronner:
i'm having this issue as well, least i think they are related. i have the following model:

```php
class products extends syncableBase
{
    protected $with=['attributes'];
    public function __construct()
    {
    }
    public function attributes()
    {
        return $this->belongsToMany('App\attributes',"productattributes","products_id","attribute_id")->withPivot(["isGroup","visibilityFormula"]);
    }
}
```

syncableBase overrides hasMany but not belongsTo and uses cacheable.

if i save an existing object (one acquired via get) there is no problems, if i save a new object without any relationships, no problems, the only problem is when i create a new object, save, sync the ids, and then save again (i don't think the separate saves are part of the issue, but if i don't do the first save it never does the sync, because there is no id for the product)

it seems the save should invalidate the cache, and given there were two saves that should doubly so

relevant part of that:

```php
$itemsArr=[];
        $groupsArr=[];
        if(isset($input["item_attributes"]))
        {
            $itemsArr=$input["item_attributes"];
        }
        if(isset($input["group_attributes"]))
        {
            $groupsArr=$input["group_attributes"];
        }
        $ids=$itemsArr+$groupsArr;
        $obj->save();//seems laravel can't handle the sync if the object is new and doesn't have a pk, as it doesn't in the case of auto_incrementing primary keys, so i have to save, then it will have a key then i can sync and save again...
        $obj->attributes()->sync($ids);
```

so after doing this, i go to a list of products as follows:

```php
$data = \App\products::orderBy('product_name', 'asc')->limit(100)->get();
```

and the product i just created isn't in the list, if i clear the cache it shows up, if i save the product i just created again (from the product edit screen) the product will show in the list also, don't let that limit bother you, you there are only 4 products in the database right now.

it's quite perplexing.

i don't THINK i'm using nova, here are my requires:
```
"require": {
        "php": "^7.1.3",
        "ckfinder/ckfinder-laravel-package": "v3.4.5.1",
        "fideloper/proxy": "^4.0",
        "genealabs/laravel-model-caching": "^0.6.1",
        "laravel/framework": "5.8.*",
        "laravel/tinker": "^1.0",
        "predis/predis": "^1.1",
        "watson/rememberable": "^3.0"
    },
    "require-dev": {
        "barryvdh/laravel-debugbar": "^3.2",
        "beyondcode/laravel-dump-server": "^1.0",
        "filp/whoops": "^2.0",
[...]

Issue #278, opened 2019-08-11
BelongsToMany Doesn't Cache When Using Laravel Nova


mikebronner:
I just want to share my experinece with this issue. I just faced the BelongsToMany relationship issue using laravel and Nova. it doesn't update redis on save or update or create using Nova. I have to flush redis to see changes made by laravel nova. But when i use disableCache() on my models, it's working fine. But almost all my models using BelongsToMany relationships. So no point to use this package if it's not going to cache my models :) It might be related with the events between nova and models? Maybe this package doesn't regionise Nova events to trigger updates on redis. 

Details about the issue:
I got Work and Service models. It's many to many relationships. Work has many Services and Services has Many Works.

Models in project looks like: 
Work Model
```php
class Work extends Model implements HasMedia
{
    use HasImage, HasTranslate, Cachable;

    public $translatable = ['client', 'project'];
    protected $collection = 'slider';

    protected $casts = [
        'date' => 'datetime',
    ];

    public function services()
    {
        return $this->belongsToMany(Service::class);
    }
}
```

Service Model
```php
class Service extends Model
{
    use HasTranslate, HasSlug, Cachable;

    public $translatable = ['name', 'description'];

    public function getSlugOptions(): SlugOptions
    {
        return SlugOptions::create()
            ->generateSlugsFrom('name')
            ->saveSlugsTo('slug');
    }

    public function works()
    {
        return $this->belongsToMany(Work::class);
    }
}
```

Works table
```php
    public function up()
    {
        Schema::create('works', function (Blueprint $table) {
            $table->bigIncrements('id');
            $table->json('client');
            $table->json('project');
            $table->date('date');
            $table->string('country');
            $table->string('url')->nullable();
            $table->timestamps();
        });
    }
```

Services table
```php
    public function up()
    {
        Schema::create('services', function (Blueprint $table) {
            $table->bigIncrements('id');
            $table->json('name');
            $table->json('description');
            $table->string('slug')->unique();
            $table->string('icon');
            $table->timestamps();
        });
    }
```

Service Work pivot table
```php
    public function up()
    {
        Schema::create('service_work', function (Blueprint $table) {
            $table->bigIncrements('id');
            $table->integer('service_id')->unsigned();
            $table->integer('work_id')->unsigned();
        });
    }
```

Laravel : 5.8.30
Redis server v=5.0.5 
Laravel Nova 2.0.9

I hope this help you to figure out the issue.  

Update: 
[...]

PR #327, opened 2020-04-09
Add nova tests

### LINKED ISSUE #283: Writing Nova Tests
If anyone can point me towards some examples of writing tests for nova packages (not apps), I'd be hugely grateful. I'm trying to set up tests for Nova, but am having trouble getting it to run.

https://github.com/GeneaLabs/laravel-model-caching/pull/284

dmason30:
Fixes #283, #278, #280 

I created this from `add-nova-tests` branch and then merged in `master` to make sure everything is up to date. 

I changed the nova support to `3.x` as that is the laravel 7 compatible version. 

This should work for you locally as an initial setup for nova I feel that either performing all actions via the api or by accessing the nova related code directly would be the best approach. 

The test is still just a placeholder for whatever actual tests need to be done.


mikebronner:
Are you submitting a new PR for this? :) Just curious why you closed it. Was going to look into it more in the near future.

dmason30:
@mikebronner Sorry, I accidentally closed it ha I have just added some initial tests for CUD and BelongsToMany and I didn't come across anything  broken but maybe I am missing something?

mikebronner:
After merging this, I am getting the following error when trying to run PHPUNIT:
```
1) GeneaLabs\LaravelModelCaching\Tests\Feature\Nova\BelongsToManyTest::testAttachRelationFlushesCache
Expected status code 200 but received 500.
Failed asserting that 200 is identical to 500.

/Users/mike/Developer/Sites/laravel-model-caching/vendor/laravel/framework/src/Illuminate/Testing/TestResponse.php:185
/Users/mike/Developer/Sites/laravel-model-caching/tests/Feature/Nova/BelongsToManyTest.php:24
```

Any suggestions?

dmason30:
@mikebronner I hgad some issues with this and found that running full suite seemed to the trick. I though committing the sqlite files would mean the table would already migrated. 

To be honest i find the setup quite confusing I understand the aim is to speed up tests but I feel neither of the sqlite files should be committed. I would remove the `AlwaysRunFirstTest` and instead migrate `baseline.sqlite` in a base `TestCase` once per run and keep the copying to `testing.sqlite`.  Then `gitignore` the sqlite files.

mikebronner:
Hi @dmason30 -- the sqlite files probably don't need to be versioned, you are right about that. And they are recreated when the tests are run. The problem with moving it to a base TestCase is that it is then recreated once for every class. Moving it out to a test that is run before all other tests was the only way to make this work only once per run.

I was running the whole suite when I got the errors above, but didn't have more time to investigate.

dmason30:
@mikebronner You can still make it run once with a TestCase just use wrap it in a static flag hence I mentioned the `self::$migrated = true`. I'll see if I have time to create a PR with a proof of concept for you.

PR #330, opened 2020-04-10
Fix broken limit test

dmason30:


dmason30:
Is there a good reason for phpunit config to have `stopOnFailure=true` ? I have always used it as false since it hides other failing tests especially on CI tools?

mikebronner:
True, but it makes it much easier for local development. If there is an error in the CI, I generally never look at all the errors, but only the fact that it errored. Then I go to my local machine and run the tests, failing at the first error. Often subsequent errors are artifacts of the first one.

Issue #333, opened 2020-04-13
Undefined offset: # in CacheKey getValuesFromBindings


irvine1231:
**Describe the bug**
I get a `Undefined offset: 6` error after the eloquent query called below.

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
Class::where('store_id', $store->id)->withAndWhereHas('enrollment', function ($query) use ($week_start, $week_end, $model_class_enrollment) {
    $query->whereIn('state', [$model_class_enrollment::STATE_PUBLISHED, $model_class_enrollment::STATE_BOOKABLE, $model_class_enrollment::STATE_STARTED, $model_class_enrollment::STATE_ENDED])->whereBetween('start_time', [$week_start->toDateTimeString(), $week_end->toDateTimeString()]);
})->get()

class Class extends Model
{
    public function scopeWithAndWhereHas($query, $relation, $constraint)
    {
        return $query->whereHas($relation, $constraint)
                    ->with([$relation => $constraint]);
    }
}
```

**Stack Trace**
```
[2020-04-13 11:52:31] local.ERROR: Undefined offset: 6 {"exception":"[object] (ErrorException(code: 0): Undefined offset: 6 at /private/var/www/project/vendor/genealabs/laravel-model-caching/src/CacheKey.php:181)
[stacktrace]
#0 /private/var/www/project/vendor/genealabs/laravel-model-caching/src/CacheKey.php(181): Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(8, 'Undefined offse...', '/private/var/ww...', 181, Array)
#1 /private/var/www/project/vendor/genealabs/laravel-model-caching/src/CacheKey.php(149): GeneaLabs\\LaravelModelCaching\\CacheKey->getValuesFromBindings(Array, '2020-04-19 15:5...')
#2 /private/var/www/project/vendor/genealabs/laravel-model-caching/src/CacheKey.php(321): GeneaLabs\\LaravelModelCaching\\CacheKey->getValuesClause(Array)
#3 /private/var/www/project/vendor/genealabs/laravel-model-caching/src/CacheKey.php(204): GeneaLabs\\LaravelModelCaching\\CacheKey->getOtherClauses(Array)
#4 [internal function]: GeneaLabs\\LaravelModelCaching\\CacheKey->GeneaLabs\\LaravelModelCaching\\{closure}('-classes_enroll...', Array)
[...]

PR #334, opened 2020-04-13
Skip adding currentBinding if the where values are not in the query binding

irvine1231:
- Skip adding currentBinding in the CacheKey class if the where values are not in the query binding.


irvine1231:
Fixes #333

PR #340, opened 2020-04-17
Fix nova install

dmason30:
It appears when you merged in the master build passed but it failed to install nova and therefore didn't run those tests hopefully this fixes it although you might just want to do some trial and error on master instead of merging this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_script:
   - pecl install pcov
   - travis_retry composer self-update;
   - travis_retry composer config --global github-protocols https;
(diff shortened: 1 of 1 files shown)


mikebronner:
Not sure if this was necessary, since the builds are passing? Probably doesn't hurt, thought.

dmason30:
@mikebronner Its still not working in master see nova is not installing:
https://travis-ci.org/github/GeneaLabs/laravel-model-caching/jobs/679088787#L445

mikebronner:
Actually, it is downloading. Look at line 448. There must be another reason why Nova tests aren't running?

dmason30:
Just noticed I accidentally left in not sure why it would cause this though:
https://github.com/GeneaLabs/laravel-model-caching/blob/master/phpunit.xml.dist#L15

mikebronner:
Doh! Forgot about php-unit.xml.dist -- thanks for finding that!

PR #343, opened 2020-04-29
~ fixes all method not recognising change to global scope params

### LINKED ISSUE #342: Global Scopes Not Being Busted Based On Called Context
**Describe the bug**
When calling Model::all() - the global scope is not being applied. I am getting an instance of the first cached result, not the updated one when the Auth::user() changed in the global scope.

In my model I attach a global scope on boot which is a class implementing scope:
```PHP
static::addGlobalScope(new SameCompanyScope);
```

However when I run the below Eloquent query, it always returns the cached results instead of updating the results when the scope changes due to a different user being logged in.

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
Model::all();
```

**Stack Trace**
There is no stack trace, because nothing "breaks" it just returns the incorrect information.

Is this expected behaviour?

The workarounds I have identified include using the get() method on the model instead of all, but this does not feel correct as the global scope should be applied to all. 

The other is modifying ModelCaching.php trait like:
```PHP
if (!$instance->isCachable() || $instance->getGlobalScopes()) {
   return parent::all($columns);
}
```

To disable the caching when a global scope are being applied, but this still feels wrong. Any thoughts or tips would be appreciated.

**Environment**
 - PHP: 7.3.5
 - OS: Windows 10 - Wamp
 - Laravel: 6.15.1
 - Model Caching: 0.7.4

Drewdan:
Fixes #342 

I have made these changes and it seems to fix the issues I raised in issue #342 

The PR adds more detail to the key, uncluding the scope bindings - this means when we reach into the cache, we take into account the global scope bindings (if any) and use this to generate a more accurate key

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -44,6 +44,7 @@ class CacheKey
         $key .= $this->getOrderByClauses();
         $key .= $this->getOffsetClause();
         $key .= $this->getLimitClause();
+        $key .= $this->getBindingsSlug();
         $key .= $keyDifferentiator;
         $key .= $this->macroKey;
 // dump($key);
@@ -351,5 +352,14 @@ class CacheKey
 
             return "{$carry}-{$relatedConnection}:{$relatedDatabase}:{$related}";
         });
+    }
+
+    protected function getBindingsSlug() : string
+    {
+        if (! method_exists($this->model, 'query')) {
+            return '';
+        }
+        
+        return Arr::query($this->model->query()->getBindings());
     }
 }
--- a/tests/Integration/CachedBuilder/ScopeTest.php
+++ b/tests/Integration/CachedBuilder/ScopeTest.php
@@ -67,7 +67,7 @@ class ScopeTest extends IntegrationTestCase
             ->first();
         $authors = (new AuthorBeginsWithScoped)
             ->get();
-        $key = sha1("genealabs:laravel-model-caching:testing:{$this->testingSqlitePath}testing.sqlite:authors:genealabslaravelmodelcachingtestsfixturesauthorbeginswithscoped-name_like_A%");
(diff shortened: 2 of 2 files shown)


mikebronner:
@Drewdan Awesome that you got started on this PR!! Thanks a bunch :) Would you mind sticking with it just a bit more to get the Travis check to pass? Thanks!

Drewdan:
Happy to help, I will take a look at the failing tests now :)

Drewdan:
@mikebronner - I cannot seem to recreate the test failure I am getting in the pipeline on travis - it does not appear locally. Are you able to shed any light on it? I am a little stuck :(

mikebronner:
Hi @Drewdan no worries about the Nova tests, they will only run on master, not on PRs.

Drewdan:
My pipeline appears to be failing because of them, is that to be expected?

Drewdan:
Hi @mikebronner - I have my pipeline failing because of some Nova tests, that said from the look of it, they should not be running:
image

I did try to modify the phpunit and travis yml files, but, I did not get much luck!

mikebronner:
Yea, no worries .. the tests will pass once integrated.

dmason30:
@mikebronner The nova tests ran in the PR because you have removed the `/** @group nova */ annotation from all the nova tests. The PR travis setup excludes the tests by using this group.

mikebronner:
@dmason30 Thanks for setting me straight! LOL I didn't realize they were there for that. I will take a look at fixing that using the test suite instead of the test group.

dmason30:
@mikebronner I experimented with that but there is no way to exclude a `testsuite`  via a command argument the only way to do it would be to add another `phpunit.xml.pr` or something but this seems extreme. Hence, why I ended up with using `@group` and using the `--exclude-group nova` argument in PRs

mikebronner:
@dmason30 I went ahead and implemented using test suites instead of groups, and got it working. But then went a step further and moved away from travis to GitHub Actions. No more issues, PRs should run fine now as well. 🤞

PR #345, opened 2020-05-02
Add PHP Matrix and caching

dmason30:
Its common to do a matrix for supported PHP versions and to cache composer dependencies for faster runs.

image

Note: Installing nova will fail on PRs as it tries to install nova when access to secrets is blocked. The same problem we had with travis.

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -31,7 +31,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php }}
-        extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, gd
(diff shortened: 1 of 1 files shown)


mikebronner:
Thanks for the updates :). Blarg, so it looks like actions don't apply env vars to PRs either ... I'll research that more this weekend.

dmason30:
@mikebronner Yeah you will probs have to go down a similar approach

```yml
- name: Remove Nova on a PR
  if: github.event_name == 'pull_request'
  run: composer remove laravel/nova --no-update --no-interaction

# Other steps...

- name: Execute Integration and Feature tests via PHPUnit
  run: vendor/bin/phpunit --configuration phpunit.xml.dist --testsuite Integration,Feature

- name: Execute Nova tests via PHPUnit
  if: github.event_name != 'pull_request'
  run: vendor/bin/phpunit --configuration phpunit.xml.dist --testsuite Nova
```

PR #346, opened 2020-05-02
Add PHP matrix and skip nova on PR

dmason30:
Its common to do a matrix for supported PHP versions and to cache composer dependencies for faster runs. And made it so they do not fail due to nova on PRs

image


mikebronner:
@dmason30 man, you're a wizard! Thanks again! :)

PR #349, opened 2020-05-03
Fix php 7.2 action

### LINKED ISSUE #348: PHP 7.2 action not actually working
https://github.com/GeneaLabs/laravel-model-caching/blob/7e7f527b0e845ab410c78324de5af7e31f61f993/.github/workflows/laravel.yml#L17

The setup php action does not support characters in the version provided from the matrix.

https://github.com/GeneaLabs/laravel-model-caching/blob/7e7f527b0e845ab410c78324de5af7e31f61f993/.github/workflows/laravel.yml#L33

This is resulting in this ⚠️ _silent_ ⚠️  error:

image
https://github.com/GeneaLabs/laravel-model-caching/runs/639471767?check_suite_focus=true#step:6:9

So this is actually running tests on 7.4 (i think).

This should be changed to 

```yml
php: [7.2, 7.3, 7.4]
```

**Note:** There is no way to force it to install a particular patch version of PHP at this time and you shouldn't need to?

dmason30:
Fixes #348

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: ['^7.2.5', 7.3, 7.4]
+        php: [7.2, 7.3, 7.4]
     
     name: PHP ${{ matrix.php }}
     


mikebronner:
Yea, this didn't work, I tried all kinds of iterations. I intentionally chose 7.2.5 as that is the minimum version it needs to run with.

mikebronner:
I was wondering how to output what version of PHP it is using. Will try to figure that out.

dmason30:
image

It does show it if you expand the Setup PHP task as long as it was provided a correct php version. See new PR #350 with the problematic dependency removed.

PR #350, opened 2020-05-03
Fix PHP 7.2 action 

### LINKED ISSUE #348: PHP 7.2 action not actually working

dmason30:
Fixes #348 

I have had to remove a dev dependency that was causing composer to fail install.

https://github.com/sebastianbergmann/phpcpd

I tried changing version to `^4.0|^5.0` but composer still failed.

If you are using this somewhere externally you will have to find a workaround for php 7.2.

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -10,17 +10,17 @@ jobs:
   laravel-tests:
 
     runs-on: ubuntu-latest
-    
+
     strategy:
       fail-fast: true
       matrix:
-        php: ['^7.2.5', 7.3, 7.4]
-    
+        php: [7.2, 7.3, 7.4]
+
     name: PHP ${{ matrix.php }}
-    
+
     steps:
     - uses: actions/checkout@v2
-      
+
     - name: Cache dependencies
       uses: actions/cache@v1
       with:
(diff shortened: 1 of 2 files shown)


dmason30:
@mikebronner Should I close this? It does fix the PHP 7.2 action but I understand you may be using the removed package externally somewhere.

mikebronner:
@dmason30 nono :) sorry, this just hasn't been a priority.

mikebronner:
I'm opting not to move forward with PHP 7.2 testing due to incompatibility.

PR #393, opened 2021-03-16
Use the part of the query (->from) instead of the table name of the model (->table)

thyseus:
Why ?

I have the following model:

```php
class DemoUnscoped {
        public $table = 'demo_unscoped';

	public function scopeFooBar($query)
	{
		$query->from('demo_scoped');

		return $query;
	}
}
```

Now, assuming my table demo_unscoped has 666 entries, and demo_scoped has 999 entries, prior to this patch i get the following results:

```
DemoUnscoped::count();
666
DemoUnscoped::FooBar()->count();
666 // <----------- WRONG (has cached the value but should not!)
```

With this fix applied i get:

```
DemoUnscoped::count();
666
DemoUnscoped::FooBar()->count();
999 // <----------- Correct ! 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.10.3] - 2021-03-16
+### Changed
+- The from part of the query ($query->from) instead of the table name of the model ($model->table)
+  is now used for determining the table part caching key fragment
+
 ## [0.10.2] - 2020-09-04
 ### Added
 - functionality to inject custom builder class for handling conflicting packages.
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -69,7 +69,7 @@ class CacheKey
 
     protected function getTableSlug() : string
     {
-        return (new Str)->slug($this->model->getTable())
+        return (new Str)->slug($this->query->from)
             . ":";
     }
 


mikebronner:
Hi @thyseus, thanks for submitting this. Unfortunately the tests aren't passing. Could you take a look into that? Be sure to run the tests locally on your machine and make sure they all pass. I'll try to take a look at it when I have some free time. Thanks! :)

thyseus:
Hey, @mikebronner  - thanks a lot for your quick reply !

I already begun digging into this issue. It seems like i need to adjust the `$key` fixtures in (every??) test so the tests are using the new source. Unfortunately i can not grasp where the "table/from" part is inside the key:

```php
	$key = sha1("genealabs:laravel-model-caching:testing:{$this->testingSqlitePath}testing.sqlite:book-store:genealabslaravelmodelcachingcachedbelongstomany-book_store.book_id_=_{$bookId}");
		$tags = [
			"genealabs:laravel-model-caching:testing:{$this->testingSqlitePath}testing.sqlite:genealabslaravelmodelcachingtestsfixturesstore",
		];
```

- which part in $key does represent the tablename ? 

Could you give me a little advice ? You know the code much better than me :)

mikebronner:
Ah, yes!!! It's annoying and tedious :) Unfortunately your change seems that it could be affecting all cache keys. Let me take a look as to why they are changing. Most tests should not really change, unless `from()` is not set in the query.
I also need to add PHP 8.0 to the test matrix.

thyseus:
Thanks a very lot ! Your module is _awesome_. Let me know if i can be of any help...

mikebronner:
@thyseus try release 0.11.2 and let me know how it works for you. :) Thanks for submitting this PR! :)

thyseus:
@mikebronner just did run `composer update`, `geneolabs/laravel-model-caching` got updated and everything is working as expected ! Thanks again a lot for this really quick reaction.

mikebronner:
@thyseus That's awesome! :) Super glad you are enjoying this package.

PR #398, opened 2021-04-13
Move slevomat/coding-standard to require-dev

dmason30:
Hi,

I believe this package was accidentally included in require instead of require-dev which means that when using this package and running `composer install --no-dev` this heavy package is still installed.

I use vapor which has size limits and this change threw us over the limit.


dmason30:
@mikebronner If you could find a moment to review this it would be much appreciated!

mikebronner:
Sorry about that. That was an unintentional oversight, for sure. Thanks for the PR :)

PR #403, opened 2021-06-24
Having clauses missing from the generated CacheKey

charlie-waddell:
The issue:

When a query has having clauses, the changes are not detected because these appear to be missed from the generation of the cache key. Example:

```
// Assume there are 100 orders, the count will be 100
Order::query()
    ->selectRaw('updated_at as aliased_updated_at')
    ->count();

// Assume there are only 50 with an updated_at value greater than the given date/time,
// the count wil be pulled from the cached result of the above query, giving us 100 instead of 50
Order::query()
    ->selectRaw('updated_at as aliased_updated_at')
    ->having('aliased_updated_at', '>' ,'2021-01-01 00:00:00')
    ->count()
```


mikebronner:
@charlie-waddell Thanks so much for taking the time to write up a PR. I will try to get this merged soon.

mikebronner (changes requested):
Thanks for submitting this! Could you write some tests to cover the code you added? Thanks!

Restingo:
@charlie-waddell did you find some time to add tests? Thank you.

mikebronner:
If anyone would like to pick this up and add tests, that would be great. I'm happy to merge once those are in.

PR #408, opened 2021-07-27
Fix pagination links not linking to correct domain for websites with multiple domains

### LINKED ISSUE #407: Pagination links are incorrectly cached when the website has multiple domains
**Describe the bug**

If you have a website which has multiple domains, and subsequently multiple sessions, you might visit a page showing paginated resources, and the class which has the pagination data, such as the path for the links is cached. This means the baseURL for the pagination links is also cached, and won't change when the URL of the website changes.

For example:
go to mywebsite.com - and load a page with paginated data - the links will show with the correct URL
process to mywebsite.app - and loads the same page with paginated data - the links will show with the previous URL, not the current one.

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
User::paginate();
```

**Stack Trace**
N.A

**Environment**
 - PHP: 7.4
 - OS: Windows 10, Ubuntu 20.04 (WSL) and AWS Llambdas
 - Laravel: 8.36.2
 - Model Caching 0.11.1

**Additional context**

This behaviour is not too unpleasant and quite edge casey, however, it does get confusing when a pagination link changes your domain from one to another, and then you lose your session.

I have a fix in mind, and the PR will follow.

Drewdan:
This MR adds a key, so caches are per domain, rather than multi domain. This prevents issue like those stated in #407 

Close #407

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -4,6 +4,7 @@ use Exception;
 use GeneaLabs\LaravelModelCaching\Traits\CachePrefixing;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\URL;
 use Illuminate\Support\Str;
 use Ramsey\Uuid\Uuid;
 
@@ -45,6 +46,7 @@ class CacheKey
         $key .= $this->getOffsetClause();
         $key .= $this->getLimitClause();
         $key .= $this->getBindingsSlug();
+        $key .= $this->getPathSlug();
         $key .= $keyDifferentiator;
         $key .= $this->macroKey;
 // dump($key);
@@ -361,5 +363,10 @@ class CacheKey
         }
 
         return Arr::query($this->model->query()->getBindings());
+    }
+
+    protected function getPathSlug() : string
+    {
+    	return ':' . strtolower(URL::to('/'));
     }
 }
--- a/tests/Feature/PaginationTest.php
+++ b/tests/Feature/PaginationTest.php
@@ -1,5 +1,9 @@
 <?php namespace GeneaLabs\LaravelModelCaching\Tests\Feature;
 
+use Illuminate\Support\Facades\App;
+use Illuminate\Support\Facades\URL;
+use Illuminate\Support\Facades\Config;
+use Illuminate\Support\Facades\Artisan;
 use GeneaLabs\LaravelModelCaching\Tests\FeatureTestCase;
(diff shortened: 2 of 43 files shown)


Drewdan:
Wow! This has a lot more tests than I remember 😅 - all the tests are passing now. I have no access to the better code hub though, so I cannot see what is wrong there. Let me know if this PR needs any changes :)

Nutickets:
I just encountered this scenario of pagination links containing the wrong domain when retrieved from cache and my first thoughts were exactly the same as your proposed solution of caching by domain.

However I think this isn't the way for 2 reasons:
1. You are increasing the amount of data in the cache by caching the same data n times depending on how many alt domains you have.
2. This doesn't consider scenarios where you pull the same paginated data from multiple different URL paths.

The second point is the bigger concern because imagine you have a page for assigning a role to multiple users and the URL is something like `https://example.com/roles/1/assign` (where 1 is a role ID). If you call `User::paginate()` on that page, then proceed to `/roles/2/assign`.. the moment you click a pagination URL, you are gonna be sent to the URL where the users were first cached, which would be `/roles/1/assign`.

It's highly likely that the user would even notice the URL change and now they are assigning role 1 instead of role 2, which could be catastrophic.

**Proposed Solution**
Assuming pagination links should always be based off the current URL (cannot think of a situation where this wouldn't hold true), it makes sense to re-calibrate the URLs upon retrieval of a paginated results set from the cache. This turns out to be actually quite a trivial fix as you can simply hook into the part where Laravel unserializes the paginator after retrieving it from Redis.

```php
// .. inside AppServiceProvider.php
$this->app->bind(LengthAwarePaginator::class, function ($app, $values) {
    return new YourCustomLengthAwarePaginator(...array_values($values));
});
```

```php
// .. inside YourCustomLengthAwarePaginator.php
public function __wakeup()
{
    if ($path = request()->getPathInfo()) {
        $this->setPath($path);
    }
}
```

Of course the above involves overriding the native Laravel paginator and really could be done automatically from the model caching library, but this solution also worked for us in places where we were applying manual caching (rather than using model cache) to paginated results so thought it was worth sharing.

Drewdan:
Hi @Nutickets 

Thank you for your comment and thoughts.

In our software solution, we are also using a similar approach you have mentioned, in that we are setting the path every time we create pagination links rather than resolving a custom paginator.

Good spot on the caching of the URL. I guess as I am just adding the root URL to the key, it would not differentiate properly between the URLs.

I think, with this in mind, will close this PR and start a fresh with this new knowledge!

Thank you :)

PR #409, opened 2021-09-02
[bug] Fix current binding count when previous binding is false

dmason30:
This bug is difficult to explain clearly so I will try my best.

I discovered it when I was trying query where a relation had a false column then using a `whereRaw` after. See this example below from the added test in this PR:

```php
$books = (new Book)
    ->whereHas('author', function ($query) {
        return $query->where('is_famous', false); // Passing false here is the issue
    })
    ->whereRaw("title = ?", ['Mixed Clause']) // This binding ends up not being included in the cache key
    ->get();
```

**Issue**
Before this PR fix the cache key that is generated would be missing the 'Mixed Clause' binding see below where the key ends without the value. 

```
-exists-and_books.author_id_=_authors.id-is_famous_=_-authors.deleted_at_null-_and_title_=
```

**Workaround**
If you use zero instead as the value `$query->where('is_famous', 0)` the key is generated correctly.

**Cause**
This is caused by false bindings being ignored in `CacheKey::getValuesFromBindings` and the `currentBindings` count was therefore not being incremented correctly. 

```php
protected function getValuesFromBindings(array $where, string $values) : string
{
    // If this array value is false then the binding is ignored
    if (($this->query->bindings["where"][$this->currentBinding] ?? false) !== false) {
        // ...
        $this-currentBinding++;
        // ...
```

**Fix**
Rather than `CacheKey::getValuesFromBindings` comparing using false it should use some specific value to compare with so it only skips if the binding is not set. Feel free to feedback on this but its a necessary evil to have some sort of unique string provided that is unlikely to be passed as a field value. 

```php
protected function getValuesFromBindings(array $where, string $values) : string
{
    // Fallback to this when the current binding does not exist in the bindings array
    $bindingFallback = __CLASS__ . ':UNKNOWN_BINDING';
    
    if (($this->query->bindings["where"][$this->currentBinding] ??  $bindingFallback) !==  $bindingFallback) {
``` 

And this ensures that the generated cache key contains all the correct bindings:
```
-exists-and_books.author_id_=_authors.id-is_famous_=_-authors.deleted_at_null-_and_title_=_Mixed_Clause
```

**Potential Additional Change?**
Not sure if you noticed but when the value is `false` the generated cache key inserts the `false` as an empty string `-is_famous_=_-` . 

I wonder if we should detect `false` values and make the key insert them as `0` or `'false'` string, e.g. `-is_famous_=_0- ` or `is_famous_=_false-`?

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -69,7 +69,7 @@ class CacheKey
         return "-{$where["boolean"]}_{$where["first"]}_{$where["operator"]}_{$where["second"]}";
     }
 
-    protected function getCurrentBinding(string $type, string $bindingFallback): string
+    protected function getCurrentBinding(string $type, mixed $bindingFallback = null): mixed
     {
         return data_get($this->query->bindings, "{$type}.{$this->currentBinding}", $bindingFallback);
     }
@@ -87,7 +87,7 @@ class CacheKey
 
         $type = strtolower($where["type"]);
         $subquery = $this->getValuesFromWhere($where);
-        $values = collect($this->getCurrentBinding('where', ""));
+        $values = collect($this->getCurrentBinding('where', []));
 
         if (Str::startsWith($subquery, $values->first())) {
             $this->currentBinding += count($where["values"]);
@@ -210,7 +210,7 @@ class CacheKey
 
         while (count($queryParts) > 1) {
             $clause .= "_" . array_shift($queryParts);
-            $clause .= $this->getCurrentBinding("where", "");
+            $clause .= $this->getCurrentBinding("where");
             $this->currentBinding++;
         }
 
@@ -282,7 +282,7 @@ class CacheKey
             $this->currentBinding++;
 
             if ($where["type"] === "between") {
-                $values .= "_" . $this->getCurrentBinding("where", "");
+                $values .= "_" . $this->getCurrentBinding("where");
                 $this->currentBinding++;
             }
         }
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -87,7 +87,7 @@ class CacheKey
 
         $type = strtolower($where["type"]);
         $subquery = $this->getValuesFromWhere($where);
-        $values = collect($this->getCurrentBinding('where', []));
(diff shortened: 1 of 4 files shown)


src/CacheKey.php
```diff
@@ -174,7 +174,10 @@ protected function getValuesFromWhere(array $where) : string
 
     protected function getValuesFromBindings(array $where, string $values) : string
     {
-        if (($this->query->bindings["where"][$this->currentBinding] ?? false) !== false) {
+        // Fallback to this when the current binding does not exist in the bindings array
+        $bindingFallback = __CLASS__ . ':UNKNOWN_BINDING';
+
+        if (($this->query->bindings["where"][$this->currentBinding] ?? $bindingFallback) !== $bindingFallback) {
```
mikebronner, line 180: Thanks for submitting this. I see the problem you are trying to solve, thanks for catching that!

I would prefer to use `data_get()` to fetch the binding, if you can update the PR to that, I'll go ahead and merge.

mikebronner (changes requested):

src/CacheKey.php
dmason30: This way of retrieving the current binding was used in multiple places in the class so I extracted it to a private function.

src/CacheKey.php
```diff
@@ -174,12 +174,16 @@ protected function getValuesFromWhere(array $where) : string
 
     protected function getValuesFromBindings(array $where, string $values) : string
     {
-        if (($this->query->bindings["where"][$this->currentBinding] ?? false) !== false) {
-            $values = $this->query->bindings["where"][$this->currentBinding];
+        // Fallback to this when the current binding does not exist in the bindings array
+        $bindingFallback = __CLASS__ . ':UNKNOWN_BINDING';
+        $currentBinding = $this->getCurrentBinding('where') ?? $bindingFallback;
```
mikebronner, line 179: Thanks for the update! What I wanted to avoid was using `?? $bindingFallback`. Let me see if I can edit it.

mikebronner:
Thank you for all the work you put into this!!

Issue #441, opened 2023-02-24
Object of class Illuminate\\Database\\Query\\Expression could not be converted to string


vheins:
**Describe the bug**

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
$data->whereHas($relation, function ($q) use ($search, $searchKey) {
	$table = $q->getModel()->getTable();
	$q->whereRaw("LOWER(cast(" . $table . "." . $searchKey . " as text)) ilike '%" . $search . "%'");
});
```

**Stack Trace**
```json
{
    "success": false,
    "code": 500,
    "status": "Internal Server Error",
    "message": {
        "error": "Object of class Illuminate\\Database\\Query\\Expression could not be converted to string",
        "trace": [
            {
                "file": "/var/www/html/vendor/genealabs/laravel-model-caching/src/CacheKey.php",
                "line": 346,
                "function": "getColumnClauses",
                "class": "GeneaLabs\\LaravelModelCaching\\CacheKey",
                "type": "->"
            },
            {
                "file": "/var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php",
                "line": 734,
                "function": "GeneaLabs\\LaravelModelCaching\\{closure}",
                "class": "GeneaLabs\\LaravelModelCaching\\CacheKey",
                "type": "->"
            },
            {
                "file": "/var/www/html/vendor/genealabs/laravel-model-caching/src/CacheKey.php",
                "line": 352,
                "function": "reduce",
                "class": "Illuminate\\Support\\Collection",
                "type": "->"
            },
            {
                "file": "/var/www/html/vendor/genealabs/laravel-model-caching/src/CacheKey.php",
                "line": 154,
                "function": "getWhereClauses",
                "class": "GeneaLabs\\LaravelModelCaching\\CacheKey",
                "type": "->"
            },
            {
                "file": "/var/www/html/vendor/genealabs/laravel-model-caching/src/CacheKey.php",
                "line": 345,
                "function": "getNestedClauses",
                "class": "GeneaLabs\\LaravelModelCaching\\CacheKey",
                "type": "->"
            },
            {
                "file": "/var/www/html/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php",
                "line": 734,
                "function": "GeneaLabs\\LaravelModelCaching\\{closure}",
                "class": "GeneaLabs\\LaravelModelCaching\\CacheKey",
[...]

PR #451, opened 2023-05-11
Add expressionToString to getOtherClauses

padre:
Fix exception: Object of class Illuminate\Database\Query\Expression could not be converted to string

Related: https://github.com/GeneaLabs/laravel-model-caching/issues/441

 Eloquent query that caused the bug, for example:

```

$this
            ->vigilances()
            ->has('investigations', '=', 1);
```

If i try to dump $where variable and got:

```
array:5 [▼ // vendor/genealabs/laravel-model-caching/src/CacheKey.php
  "type" => "Basic"
  "column" => Illuminate\Database\Query\Expression {#4958 ▼
    #value: "(select count(*) from `investigations` inner join `investigation_vigilance` on `investigations`.`id` = `investigation_vigilance`.`investigation_id` where `vigilances`.`id` = `investigation_vigilance`.`vigilance_id` and `investigations`.`deleted_at` is null) ◀"
  }
  "operator" => "="
  "value" => Illuminate\Database\Query\Expression {#4960 ▼
    #value: 1
  }
  "boolean" => "and"
]
```

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -207,8 +207,8 @@ class CacheKey
 
         $column = "";
 
-	if (isset($where["column"]) && $where["column"] instanceof Expression) {
-            $where["column"] = $this->expressionToString($where["column"]);
+	if (data_get($where, "column") instanceof Expression) {
+            $where["column"] = $this->expressionToString(data_get($where, "column"));
         }    
 
         $column .= isset($where["column"]) ? $where["column"] : "";
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -206,6 +206,11 @@ class CacheKey
         $value .= $this->getValuesClause($where);
 
         $column = "";
+
+	if (isset($where["column"]) && $where["column"] instanceof Expression) {
+            $where["column"] = $this->expressionToString($where["column"]);
+        }    
+
         $column .= isset($where["column"]) ? $where["column"] : "";
         $column .= isset($where["columns"]) ? implode("-", $where["columns"]) : "";
 


padre:
@mikebronner can you merge this, please?

Thanks!

mikebronner:
@padre could you add a test that covers this specific use-case?

padre:
@mikebronner Thank you very much for your quick response and your attention, but I don't think I have the necessary knowledge to add a test.

The exception occurs when I use ->has() in a belongsToMany relationship.

```class Investigation extends Model
{
    public function vigilances()
    {
       return $this->belongsToMany(Vigilance::class);
    }
}```

```class Vigilance extends Model
{
    public function investigations()
    {
        return $this->belongsToMany(Investigation::class);
    }
}```

```
$investigation
->vigilances()
->has('investigations', '=', 1)
->get();
```

This causes in the method "getOtherClauses(array $where)" in the statement

```$column .= isset($where["column"]) ? $where["column"] : "";```

 the exception "Object of class IlluminateDatabaseQueryExpression could not be converted to string".

If i try to dump $where variable and got:


```
array:5 [
  "type" => "Basic"
  "column" => Illuminate\Database\Query\Expression {#3051 ▼
    #value: "(select count(*) from `investigations` inner join `investigation_vigilance` on `investigations`.`id` = `investigation_vigilance`.`investigation_id` where `vigilances`.`id` = `investigation_vigilance`.`vigilance_id` and `investigations`.`deleted_at` is null) ◀"
  }
  "operator" => "="
  "value" => Illuminate\Database\Query\Expression {#3053 ▼
    #value: 1
  }
  "boolean" => "and"
```

Thank you very much for your dedication.

PR #456, opened 2023-11-10
Update CacheKey.php to allow null relationships

peter-brennan:
Allow null relationships


Flobbos:
Will this be accepted soon? I think it makes a lot of sense and I really need it. :-)

Issue #437, opened 2022-11-18
Saving not working with eloquent strict mode


jamesRUS52:
When using  in app service provider
```php
Model::shouldBeStrict(!$this->app->isProduction());
```
saving model not working with exception
Illuminate \ Database \ Eloquent \ MissingAttributeException
The attribute [query] either does not exist or was not retrieved for model [app\Models\MyModel].

**Eloquent Query**
Please provide the complete eloquent query that caused the bug, for example:
```php
$record = MyModel::find($id);
$record->field = 'foo';
$record->save();
```

**Stack Trace**
Illuminate \ Database \ Eloquent \ MissingAttributeException
The attribute [query] either does not exist or was not retrieved for model [app\Models\MyModel].

**Environment**
PHP 8.1.6
Laravel 9.36.3
Model Caching: 0.12.5

** work around **
add to model
```php
protected $query = null; // required property for Cachable trait
```
I think you should add this in trait

jamesRUS52:
up

Adesin-fr:
Also facing this !

mikebronner:
This should be fixed now in version 12.0.0. Closing this issue, please open a new issue if this is still a problem.

PR #465, opened 2024-04-16
Fixes `query` attribute not existing under strict mode

daika7ana:
When saving a model and `Model::shouldBeStrict()` is enabled, or to be more precise `Model::preventAccessingMissingAttributes()`, the `$query` attribute will not be accessible with the following error:
```
The attribute [query] either does not exist or was not retrieved for model [App\Models\ExampleModel].
{"userId":1,"exception":"[object] (Illuminate\\Database\\Eloquent\\MissingAttributeException(code: 0): 
The attribute [query] either does not exist or was not retrieved for model [App\\Models\\ExampleModel]. 
at /app/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php:475)
```

This PR should fix this issue without impacting any functionality. Will solve #437.

### DIFF
--- a/src/Traits/ModelCaching.php
+++ b/src/Traits/ModelCaching.php
@@ -22,6 +22,11 @@ trait ModelCaching
                 ?? 0;
         }
 
+        if ($key === 'query') {
+            return $this->query
+                ?? $this->newModelQuery();
+        }
+
         return parent::__get($key);
     }
 


SkyLundy:
Thanks for this fix. Solved the issue for me.

PR #466, opened 2024-04-25
Fix: Handle of DateTime objects in whereIn and whereBetween

roncodes:
Currently the `CacheKey` breaks when a DateTime instance is provided in where clauses like `whereIn`, `whereBetween`, this PR makes changes to `getCurrentBinding` and `processEnum` to handle `DateTime` instances when they are passed to where `values` array


mikebronner:
@roncodes Thanks for the PR! I will review and merge this soon.

mikebronner:
tests are failing, need to get those to pass

PR #469, opened 2024-06-16
[Fix] Carbon v3 new method default parameter compatibility

daika7ana:
Takes into account Carbon v3 new default parameter value on the `diffInSeconds` method.

Carbon v2 method blueprint:
```
public function diffInSeconds($date = null, $absolute = true)
```

Carbon v3 method blueprint:
```
public function diffInSeconds($date = null, bool $absolute = false): float
```

Fixes failing tests:
https://github.com/mikebronner/laravel-model-caching/blob/a38c905790c73c43dac09790eec7f0972cd651cc/tests/Integration/CachedModelTest.php#L180

### DIFF
--- a/src/Traits/Caching.php
+++ b/src/Traits/Caching.php
@@ -241,7 +241,7 @@ trait Caching
 
         if (
             ! $cacheCooldown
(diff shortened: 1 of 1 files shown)


mikebronner:
Needs to run and pass all tests, for some reason they weren't triggered.

PR #477, opened 2025-02-07
fix deprecations with nullable func args in php 8.4

### LINKED ISSUE #475: Fix Deprecation message for php 8.4
"
PHP message: PHP Deprecated:  GeneaLabs\LaravelModelCaching\Traits\ModelCaching::scopeWithCacheCooldownSeconds(): Implicitly marking parameter $seconds as nullable is deprecated, the explicit nullable type must be used instead in /var/www/app/vendor/genealabs/laravel-model-caching/src/Traits/ModelCaching.php on line 161; 
"

"
PHP message: PHP Deprecated:  GeneaLabs\LaravelModelCaching\Traits\BuilderCaching::withoutGlobalScopes(): Implicitly marking parameter $scopes as nullable is deprecated, the explicit nullable type must be used instead in /var/www/app/vendor/genealabs/laravel-model-caching/src/Traits/BuilderCaching.php on line 34;
"

d8vjork:
Fixes #475

### DIFF
--- a/src/Traits/BuilderCaching.php
+++ b/src/Traits/BuilderCaching.php
@@ -31,7 +31,7 @@ trait BuilderCaching
         return parent::withoutGlobalScope($scope);
     }
 
-    public function withoutGlobalScopes(array $scopes = null)
+    public function withoutGlobalScopes(?array $scopes = null)
     {
         if ($scopes !== null) {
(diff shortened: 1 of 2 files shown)


d8vjork:
Probably should add PHP 8.4 to the tests matrix to ensure this isn't breaking again

PR #517, opened 2025-04-28
Added failing test for caching query with % character in value string

JkappenFlashpoint:
https://github.com/mikebronner/laravel-model-caching/issues/516


mikebronner:
implemented in #521 
thanks for providing the tests!!

PR #519, opened 2025-10-24
Allow using custom builders on models

akyrey:
Use original model newEloquentBuilder to retrieve the builder used for a model.
This allows using custom builders on models, defining them with UseEloquentBuilder attribute or $builder static property


vipex:
@mikebronner any idea on how would it take to release this one? (if any)

mikebronner:
Finally found a way to do this. #528 will be released as a pre-release

PR #564, opened 2026-02-25
fix(cache): Cachable trait interferes with dynamic $fillable changes

### LINKED ISSUE #534: Cachable trait interferes with dynamic $fillable changes in model constructor
## Summary

When a model uses the `Cachable` trait and dynamically modifies `$fillable` in its `__construct()` method (e.g. based on the authenticated user's role), the fillable change requires **two saves** to take effect. Removing the `Cachable` trait resolves the issue.

## Reproduction

```php
class MyModel extends Model
{
    use Cachable;

    public function __construct(array $attributes = []) {
        if (auth()->user()->is_admin) {
            $this->fillable[] = 'is_active';
            $this->fillable[] = 'sales_discount';
        }
        parent::__construct($attributes);
    }
}
```

1. Log in as an admin.
2. Update `is_active` on an instance of `MyModel`.
3. **First save is silently ignored.**
4. Second save applies correctly.

**Without** `Cachable` trait: first save works correctly.

## Environment

- PHP: 7.4
- OS: Ubuntu 18.04
- Laravel: 6.18
- Model Caching: 0.7.4

## Notes

The likely cause is that the `Cachable` trait interferes with constructor-time model setup, possibly caching a stale model state or interfering with `__construct` execution order.

_Converted from discussion [#486](https://github.com/GeneaLabs/laravel-model-caching/discussions/486)._

### Acceptance Criteria

- [x] Dynamically modifying `$fillable` in a model's `__construct()` works correctly on the first save when the `Cachable` trait is applied.
- [x] Removing `Cachable` does not change the observed behavior (i.e. both behave identically).

### Test Coverage

- [x] Unit test: model with `Cachable` + dynamic `$fillable` in constructor — assert field is saved on first attempt.
- [x] Regression test: ensure constructor-set fillable fields are not lost/overwritten by cache-related initialization.
---

## WSJF Priority Score
- **Score:** 3.20
- **User/Business Value:** 6 (Dynamic fillable in constructor broken with Cachable)
- **Time Criticality:** 5 (Constructor modification is a documented Laravel pattern)
- **Risk Reduction/Opportunity Enablement:** 5 (Prevents silent double-save bug)
- **Job Size:** 5 points (S)
- **Calculation:** (6 + 5 + 5) ÷ 5 = 3.20

mikebronner:
## Summary

When a model uses the `Cachable` trait and dynamically modifies `$fillable` in its `__construct()` (e.g., based on auth context), cached models deserialized in a different context have stale `$fillable` values because PHP's `unserialize()` does not call `__construct()`. This causes mass assignment to silently ignore fields that should be fillable — the "first save silently ignored" bug.

The fix adds a `__wakeup()` method to the `Cachable` trait that re-runs the model constructor on deserialization, ensuring dynamic property modifications reflect the current request context. Volatile instance state (attributes, original, relations, etc.) is preserved around the constructor call.

## Acceptance Criteria

- [x] Dynamically modifying `$fillable` in a model's `__construct()` works correctly on the first save when the `Cachable` trait is applied.
- [x] Removing `Cachable` does not change the observed behavior (i.e. both behave identically).
- [x] Unit test: model with `Cachable` + dynamic `$fillable` in constructor — assert field is saved on first attempt.
- [x] Regression test: ensure constructor-set fillable fields are not lost/overwritten by cache-related initialization.

Fixes #534

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -6,7 +6,6 @@ namespace GeneaLabs\LaravelModelCaching;
 
 use BackedEnum;
 use DateTimeInterface;
-use Exception;
 use GeneaLabs\LaravelModelCaching\Traits\CachePrefixing;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Support\Arr;
@@ -141,13 +140,17 @@ class CacheKey
         $type = strtolower($where["type"]);
         $subquery = $this->getValuesFromWhere($where);
 
-        if (! is_numeric($subquery) && ! is_numeric(str_replace("_", "", $subquery))) {
+        if (
+            ! is_numeric($subquery)
+            && ! is_numeric(str_replace("_", "", $subquery))
+            && strlen($subquery) === 16
+        ) {
             try {
                 $subquery = Uuid::fromBytes($subquery);
                 $values = $this->recursiveImplode([$subquery], "_");
 
                 return "-{$where["column"]}_{$type}{$values}";
-            } catch (Exception $exception) {
+            } catch (Throwable) {
                 // do nothing
             }
         }
--- a/tests/Integration/CachedBuilder/MorphOneTest.php
+++ b/tests/Integration/CachedBuilder/MorphOneTest.php
@@ -44,10 +44,10 @@ class MorphOneTest extends IntegrationTestCase
             ->tags($tags)
             ->get($key2)['value'];
 
-        $this->assertEquals($cachedResults1->pluck("images.id"), $books1->pluck("images.id"));
-        $this->assertEquals($cachedResults2->pluck("images.id"), $books2->pluck("images.id"));
-        $this->assertNotEquals($cachedResults1->pluck("images.id"), $cachedResults2->pluck("images.id"));
-        $this->assertNotEquals($books1->pluck("images.id"), $books2->pluck("images.id"));
+        $this->assertEquals($cachedResults1->pluck("image.id"), $books1->pluck("image.id"));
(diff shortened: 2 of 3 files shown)


mikebronner (commented):
## Review Complete ✅

The `__wakeup()` approach is well-considered — re-running the constructor while preserving deserialized state handles the context-switch case cleanly. The serialization round-trip test with admin/non-admin context is particularly strong. Test coverage is thorough.

mikebronner (commented):
## Review — PR #564

**AC Verification:**
- ✅ `__wakeup()` re-runs constructor on deserialization, restoring dynamic `$fillable` to current context
- ✅ Volatile state (attributes, original, relations, exists, connection) preserved across constructor call
- ✅ Tests verify: first save works, matches uncached behavior, context-dependent fillable survives serialization round-trip
- ✅ Regression test: cached in non-admin context, deserialized in admin context — dynamic fillable reflects admin context

**Code Review:**
- The `__wakeup()` approach is the correct PHP mechanism for post-deserialization hooks
- State preservation is thorough — attributes, original, relations, exists, connection all backed up and restored
- `AuthorWithConditionalFillable` with static `$adminMode` toggle is a clever test fixture for context-dependent behavior
- 7 tests with clear assertions

**One concern:**

- [ ] **`__wakeup()` calls `$this->__construct()`** — this assumes all model constructors are safe to call twice and accept zero arguments. Most Eloquent models call `parent::__construct($attributes)` which is fine with no args, but models with required constructor parameters (e.g., constructor injection) will break. Consider adding a guard: only call `__construct()` if it accepts zero required parameters, or document this limitation.

Otherwise solid work.

PR #586, opened 2026-03-31
Add DynamoDB cache store support

dgtlss:
## Summary

This PR adds DynamoDB support to `laravel-model-caching` without changing the existing behavior for currently supported taggable cache stores.

The package keeps the native tagged-cache path for existing supported stores and adds a DynamoDB-specific invalidation strategy based on versioned namespaces. This allows targeted and global invalidation without relying on native cache tags or destructive store-wide flushes.

## Why This Change Is Needed

Laravel supports DynamoDB as a cache store, but this package currently relies on native cache tags for invalidation. DynamoDB does not support that invalidation model in the same way as Redis or Memcached, which means the package cannot safely support DynamoDB by reusing the existing tag-flush path.

This PR adds DynamoDB support in a way that:
- preserves current behavior for supported taggable stores
- avoids table scans or destructive flushes in DynamoDB
- keeps the public package API unchanged

## Implementation Approach

For existing taggable stores, the package continues to use the current native `tags(...)` behavior.

For DynamoDB only, the package now uses versioned namespaces:
- one global namespace token for package-wide invalidation
- one namespace token per cache tag for targeted invalidation
- cache keys are built from the original query key plus the active namespace tokens
- invalidation works by bumping version tokens instead of deleting every matching cache row

This approach was chosen because it is the safest DynamoDB-compatible invalidation model:
- it avoids expensive scans and mass deletes
- it keeps invalidation targeted
- it preserves existing package behavior everywhere outside the DynamoDB path

Old DynamoDB rows are intentionally left to expire via TTL.

## What Changed

- added a package-internal cache repository/adapter for package-managed cache operations
- preserved the existing native tag flow for current supported stores
- added a DynamoDB-specific versioned-key strategy for cache reads, writes, and invalidation
- updated `modelCache:clear` so DynamoDB performs logical invalidation rather than store-wide `flush()`
- updated targeted invalidation paths to bump namespace tokens on DynamoDB
- kept cooldown/control keys unversioned
- extended fallback handling so DynamoDB connection failures can use `fallback-to-database`
- added DynamoDB-specific tests
- updated documentation to describe DynamoDB support and TTL expectations

## Backward Compatibility

This PR does not change the public package API.

The following interfaces remain unchanged:
- `MODEL_CACHE_STORE`
- `ModelCache::invalidate(...)`
- `php artisan modelCache:clear`

Existing supported taggable stores continue to use the original cache-tag behavior.

## Performance Notes

The DynamoDB implementation avoids destructive invalidation and avoids table scans. Invalidation becomes a lightweight namespace version bump, which is a better fit for DynamoDB’s access model and keeps reads/writes predictable.

### DIFF
--- a/.github/workflows/laravel.yml
+++ b/.github/workflows/laravel.yml
@@ -10,6 +10,11 @@ jobs:
   laravel-tests:
 
     runs-on: ubuntu-latest
+    env:
+      PGSQL_HOST: 127.0.0.1
+      PGSQL_DATABASE: testing
+      PGSQL_USERNAME: forge
+      PGSQL_PASSWORD: secret
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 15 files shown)


mikebronner:
@dgtlss thank you very much for submitting this! I will review and provide feedback.

mikebronner:
📋 **Refinement Issue Created**: #587

This PR has been reviewed for post-merge work. A companion refinement issue has been created to capture:
- Integration testing across all caching traits
- Edge case handling (version key proliferation, concurrent invalidation, TTL strategy)
- Real DynamoDB CI integration
- Documentation completeness (setup guide, troubleshooting)
- Performance baseline validation

See #587 for details and next steps.

mikebronner:
## PR #586 Review — DynamoDB Cache Store Support

Reviewed with consideration for refinement gaps in #587.

---

### Code Quality

**1. Missing type hint on `$repository` (ModelCacheRepository.php:19)**

```php
protected $repository,
```

This accepts any cache repository but has no type declaration. Should be typed as `\Illuminate\Contracts\Cache\Repository` (or `\Illuminate\Cache\Repository`). Without it, static analysis tools miss method-call errors and IDE autocompletion is blind.

**2. `ModelCacheRepository::make()` creates a new instance on every call**

Every cache operation in the traits calls `$this->modelCacheRepository()` which calls `ModelCacheRepository::make()`, which resolves the container, reads config, and instantiates the store fresh each time. Within a single request, a query-heavy page could invoke this dozens of times. Consider caching the instance on the trait (or as a singleton).

**3. `rememberForever()` silent failure path**

In `currentVersion()` (line ~136):

```php
return (string) $this->repository->rememberForever(
    $versionKey,
    fn () => $this->freshVersion(),
);
```

If the DynamoDB store is unreachable, this throws — but the callers (`versionedKey` → `itemKey` → `get`/`rememberForever`/etc.) don't catch it. The `withCacheFallback` wrapper in the traits handles this for query-level operations, but `invalidateTags()` and `invalidateAll()` are called **outside** `withCacheFallback` in several places (e.g., `Buildable::decrement`, `Buildable::delete`, `flushRelationshipCache`). A DynamoDB outage during a write/delete operation would throw an unhandled exception.

**4. Version key proliferation — no cleanup strategy**

Each `invalidateTags()` call writes a new UUID version but the **old versioned cache items** remain in DynamoDB until TTL expires. The README mentions this but doesn't specify a recommended TTL. With high-invalidation models, the table could accumulate significant dead items. Worth documenting a recommended TTL range (e.g., "set DynamoDB TTL to match your longest acceptable staleness window, typically 24-72 hours").

**5. Atomicity of version bumps**

`invalidateTags()` iterates tags and writes each version key sequentially via `forever()`. If the process dies mid-loop (e.g., 3 of 5 tags updated), you get partial invalidation — some query namespaces rotate, others don't. This is an inherent limitation worth documenting rather than fixing, but it should be called out.

---

### Test Coverage

**What's covered well:**
- Basic query caching on DynamoDB (`test_repeated_queries_reuse_cached_results`)
- Selective model invalidation (`test_model_invalidation_refreshes_only_the_target_model_namespace`)
- Global invalidation via artisan command (`test_clear_command_logically_invalidates`)
- Cooldown metadata stays unversioned
- Pivot operations
- DynamoDB connection fallback + non-connection exception propagation

**Gaps identified:**

1. **No test for `CachedValueRetrievable` or `CachesOneOrManyThrough` traits on DynamoDB.** These traits have their own `retrieveCachedValue` implementations that now call `getModelCacheValue`/`putModelCacheValue`. Need tests for HasManyThrough/HasOneThrough relations with DynamoDB store.

2. **No hash collision test on DynamoDB.** `preventHashCollision()` in both `Buildable` and `CachedValueRetrievable` calls `forgetModelCacheValue` with `hash=true`. This path needs a DynamoDB-specific test to verify the versioned+hashed key is correctly forgotten and re-cached.

3. **No concurrent invalidation test.** Two processes invalidating the same tag simultaneously — does the last-write-wins behavior produce correct results?

4. **No test for `ModelCaching::all()` on DynamoDB.** The `all()` method in `ModelCaching` trait now calls `rememberModelCacheForever` directly (not through `cachedValue`). This path needs coverage.

5. **`FakeDynamoDbStore` doesn't extend `DynamoDbStore`.** The `ModelCacheRepository::make()` detection checks `$repository->getStore() instanceof DynamoDbStore`. The fake implements `Store` only, so detection in tests relies entirely on the config driver string `'dynamodb'`. This means the `instanceof` fallback path (lines 39-42 of ModelCacheRepository) is **never tested**.

6. **No test for tag strings >100 chars or special characters.** The `tagVersionKey()` method sha1-hashes tags, which handles length — but worth a smoke test.

7. **Missing: real DynamoDB validation.** All tests use `FakeDynamoDbStore`. A LocalStack-based CI test would catch serialization issues, item size limits, and TTL behavior that the fake can't surface. (Acknowledged as a future step in #587.)

---

### Backward Compatibility

**Looks solid.** The `repositoryFor()` method correctly delegates to native tags for `TaggableStore` subclasses and only applies versioning for DynamoDB. The conditional logic is clean:

```php
if (! $this->usesDynamoDb && is_subclass_of($this->repository->getStore(), TaggableStore::class)) {
    return $this->repository->tags($tags);
}
```

Existing Redis/Memcached stores will hit the tags path as before. The trait methods (`rememberModelCacheForever`, `getModelCacheValue`, etc.) are additive — they don't change signatures of existing public methods.

**One concern:** The `cache()` method on the model (used for cooldown metadata) is still called directly in `getModelCacheCooldown` / `getCacheCooldownDetails`. This is intentional (cooldown keys are unversioned), but it means cooldown operations on DynamoDB bypass the repository and hit the raw store. If DynamoDB doesn't support the `tags()` call in `cache()`, this could break. The cooldown test passes because `FakeDynamoDbStore` doesn't enforce this — worth verifying against real DynamoDB.

---

### Documentation

The README addition is a good start but incomplete per #587:

1. **Missing: complete `cache.php` config example.** Users need to see the full store definition with `driver`, `table`, `key_attribute`, `value_attribute`, `expires_attribute`, `endpoint`, `region`, and `credentials`.

2. **Missing: TTL strategy guidance.** "Enable DynamoDB TTL" is necessary but not sufficient. Recommend a TTL value and explain the tradeoff (shorter TTL = faster cleanup but more DynamoDB writes; longer TTL = more dead items but fewer writes).

3. **Missing: "When to use DynamoDB vs Redis" section.** Users need guidance on when DynamoDB caching makes sense (serverless/Lambda, no ElastiCache, multi-region) vs. when Redis is better (low-latency, high-throughput, tag-native).

4. **Missing: troubleshooting section.** Common issues: stale data after invalidation (TTL hasn't expired yet), version key accumulation, `modelCache:clear` doesn't physically delete items.

---

### Summary

The architecture is sound — versioned namespace invalidation is the right pattern for a non-taggable store. The implementation is clean and the trait integration is well-executed. The main concerns are:

- **Type the `$repository` property**
- **Cache the `ModelCacheRepository` instance per-request** (perf)
- **Wrap `invalidateTags`/`invalidateAll` calls in `withCacheFallback`** where they're invoked during write operations
- **Expand test coverage** for through-relations, hash collision on DynamoDB, and the `instanceof DynamoDbStore` detection path
- **Complete documentation** per #587 gaps

Ref: #587 for full refinement checklist.

dgtlss:
I'll take a look now and update the PR 👀

mikebronner:
> I'll take a look now and update the PR 👀

Awesome, if you have any questions at all, hit me up. Happy to work together to get them sorted out. Or if its all too much, happy to merge and implement the changes on my end.

mikebronner:
@dgtlss just fyi: whenever you are ready, go ahead and request re-review by clicking the arrows next to my name in the reviewers list (just so I don't step on your toes):

Screenshot 2026-04-01 at 1 35 17 PM

mikebronner:
Sorry for the delay, I am just now able to get back to this.

mikebronner:
@dgtlss awesome work, thank you for all this!

PR #590, opened 2026-04-15
fix: support tpetry/laravel-postgresql-enhanced and custom expressions

0x15f:
Many packages such as tpetry/laravel-postgresql-enhanced add custom Expression classes that can be used for `where`, `whereIn`, etc. Many of these use the Contract and will fail when combined with model caching. We've had a class-override in our project for a few years that I just found when upgrading the package so opening pr for it :)


0x15f:
cc @mikebronner

mikebronner:
thanks for the nudge @0x15f ! I'll get on this in the next couple of days

PR #605, opened 2026-07-17
Fix cache keys for deferred `ofMany` eager-load queries

### LINKED ISSUE #604: Cached `ofMany` eager-loaded relations return null on later paginated pages
# `ofMany` eager-loaded relations return null on later paginated pages after cache warmup

**Describe the bug**

When a paginated Eloquent query eager-loads a cached `HasOne::ofMany()` relation, the relation is correct on the first page but may be `null` on subsequent pages after the related-model cache has been warmed.

Laravel builds the aggregate query used by `ofMany()` lazily through a `beforeQuery` callback. During eager loading, Laravel adds the current page's parent IDs to the joined subquery immediately before executing it.

Laravel Model Caching generated the related-model cache key before that callback ran. Consequently, the key did not include the joined subquery or its page-specific bindings. Different pages could therefore receive the same related-model cache key.

Page two would then retrieve page one's related models from the cache. Eloquent could not match those models to the parents on page two, so it initialized the eager-loaded relation as `null`.

This also affects the convenience methods `oldestOfMany()` and `latestOfMany()` because they use the same underlying query behavior.

**Eloquent Query**

Example relationship:

```php
use Illuminate\Database\Eloquent\Relations\HasOne;

public function oldestBook(): HasOne
{
    return $this->hasOne(Book::class)
        ->ofMany('id', 'min');
}
```

Reproduction:

```php
$pageOne = Author::query()
    ->with('oldestBook')
    ->orderBy('name')
    ->paginate(
        perPage: 1,
        page: 1,
    );

$pageTwo = Author::query()
    ->with('oldestBook')
    ->orderBy('name')
    ->paginate(
        perPage: 1,
        page: 2,
    );

// Correct:
$pageOne->first()->oldestBook;

// Incorrect after the relation cache was populated by page one:
$pageTwo->first()->oldestBook; // null
```

**Stack Trace**

There is no stack trace. The query completes successfully but returns incorrect cached relationship data. The visible symptom is a silently missing (`null`) eager-loaded relation on later pages.

**Environment**

Environment in which the behavior was reproduced:

- PHP: 8.5.8
- OS: macOS 26.5.2
- Laravel: 13.19.0
- Model Caching: 13.1.6
- Cache store: Redis
- Database: SQLite

The issue is caused by Laravel's deferred `ofMany()` query construction rather than by a particular database engine.

**Additional context**

Initial investigation indicates that the related query's cache key is created before Laravel calls its registered `beforeQuery` callbacks. At that point, neither the `ofMany()` joined-subquery SQL nor its parent-ID bindings are present in the query used to generate the key.

Additionally, the existing bindings slug appears to be derived from a fresh model query rather than from the related query being cached. It therefore cannot distinguish the eager-load queries for different pages.

A possible solution would be to incorporate the materialized deferred query into cache-key generation. This would likely require:

1. Detect related queries with deferred `beforeQuery` callbacks.
2. Clone the underlying query builder so cache-key generation does not mutate the query that Laravel will execute.
3. Apply the clone's callbacks to materialize the deferred joins.
4. Hash the materialized SQL and bindings.
5. Append a suffix to the existing cache key.

Queries without deferred callbacks should retain their previous cache-key format to avoid unnecessarily invalidating ordinary cache entries.

ezp-matt:
## Summary

Fixes #604.

Laravel builds the aggregate subquery used by `HasOne::ofMany()` through a deferred `beforeQuery` callback. Laravel Model Caching previously generated the related-model cache key before that callback ran, so the key did not include the materialized join or its page-specific bindings.

This caused eager-loaded `ofMany()`, `oldestOfMany()`, and `latestOfMany()` relations on different pagination pages to share a cache entry. Later pages could receive related models belonging to the first page, which Eloquent could not match to their parent models, leaving the relations `null`.

## Changes

- Detect queries with deferred `beforeQuery` callbacks.
- Clone the query before applying those callbacks, leaving the original query untouched.
- Materialize the callbacks on the cloned query before generating the cache key.
- Append a hash of the materialized SQL and bindings to distinguish queries with different deferred state.
- Preserve the existing cache-key format for queries without deferred callbacks.
- Add regression coverage for an eager-loaded `ofMany()` relation across pagination pages.

The regression test disables model caching on the parent query while leaving the related model cacheable. This ensures the test exercises the relation cache key directly instead of passing because pagination caching was bypassed.

### DIFF
--- a/src/CacheKey.php
+++ b/src/CacheKey.php
@@ -24,6 +24,7 @@ class CacheKey
     protected $macroKey;
     protected $model;
     protected $query;
+    protected bool $hasBeforeQueryCallbacks;
     protected $withoutGlobalScopes = [];
     protected $withoutAllGlobalScopes = false;
 
@@ -35,6 +36,13 @@ class CacheKey
         array $withoutGlobalScopes,
         $withoutAllGlobalScopes
     ) {
+        $this->hasBeforeQueryCallbacks = $query->beforeQueryCallbacks !== [];
+
+        if ($this->hasBeforeQueryCallbacks) {
+            $query = clone $query;
+            $query->applyBeforeQueryCallbacks();
+        }
+
         $this->eagerLoad = $eagerLoad;
         $this->macroKey = $macroKey;
         $this->model = $model;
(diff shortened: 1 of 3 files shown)


mikebronner:
@ezp-matt thank you for your contribution! I did end up writing this slightly differently and more encompassing in my own PR, but thank you nonetheless for surfacing the issue and taking a crack at fixing it. :)

PR #629, opened 2026-09-16
Build a cache key for a where() value of any type

bastianschwarz:
`CacheKey::processEnum()` raised a `TypeError` for any `where()` value outside the union it declared, so a query this package only means to be caching died while its key was built.

## The problem

A `where()` value is whatever the application passed, so the declared union cannot be complete. `whereJsonContains()` reaches it most plainly: it keeps the value it is given and encodes it only when the statement is compiled, so a `Collection` arrives intact.

```
TypeError: CacheKey::processEnum(): Argument #1 ($value) must be of type
BackedEnum|UnitEnum|Expression|DateTimeInterface|string|int|float|bool|null,
Illuminate\Support\Collection given
```

We hit this in a production application, where an attribute cast handed its own object to a `where()`.

## The change

`processEnum()` now takes `mixed`, and an object with no string form goes to a new `hashOpaqueObject()`. PHP implements `Stringable` automatically for any class declaring `__toString()`, so `! $value instanceof Stringable` covers the rest.

Every value that matched a branch before still matches it, so no existing key changes.

## Why hash the object rather than name its class

The segment has to tell two values apart, because a `where()` value decides which rows come back. Two `Collection`s holding different ids would share a key under `get_class()`, and one query would be served the other's rows. `serialize()` reads private state, so it separates them, and it is stable across requests.

`normalizeForHash()` nearby uses `get_class()` correctly: it feeds one hash of builder properties, where a collision costs precision rather than correctness.

## The fallback, and what it costs

`serialize()` throws on a `Closure` and on anything holding a PDO connection. The fallback is `spl_object_hash()`, which never repeats, so such a query misses its own entry and writes one nothing reads. Giving every unserializable object of a class the same key would return wrong rows instead, so the miss is the better failure.

Not caching such a query at all would be cleaner, but `CacheKey` cannot signal that; it belongs in `Buildable`/`Caching`. Happy to follow up if you want it.

## Testing

Added `WhereJsonContainsTest::testWithInUsingCollectionQueryWithCollectionValues`, next to the existing array-value test. No new fixtures. The `Collection` carries the same JSON as the array form and binds the same string, so it must produce the same key.

Reverted the fix and confirmed the test goes red with the `TypeError` above, then restored it.

After: `Tests: 544, Assertions: 1336, Skipped: 4`

Skips are the memcached extension and three DynamoDB tests needing `DYNAMODB_ENDPOINT`.

Our local patch also walked arrays in `processEnum()`. `getValuesFromWhere()` flattens first, so no array can reach it today, and I left the untested branch out.


bastianschwarz:
And yes, I already told it to be "more concise"

mikebronner (changes requested):
👋 Thank you for this. The diagnosis is right, the fix addresses a real crash, and the write-up is unusually good. You explained *why* you hashed rather than named the class. You named the fallback's cost instead of hiding it. You left out the array branch you could not test, rather than shipping it untested. That is exactly the reasoning I want to see in this package.

I am requesting changes on four items. One is a correctness blocker. The other three are smaller, and I would like all four addressed before this lands.

---

## 🔴 1. `spl_object_hash()` repeats, so the fallback can serve the wrong rows

This is the blocker. The comment on `hashOpaqueObject()` states:

> the fallback is unique per instance and never repeats: the key misses instead of returning rows that belong to a different query

PHP recycles an object handle as soon as the object is destroyed. `spl_object_hash()` is derived from that handle, so it repeats often. Running your function verbatim on PHP 8.5.8:

```
value A ids=[1,2,3]  -> OpaqueWhereValue-f98ad7d2a26d333e9d25c1bf791b80754ae1acd0
value B ids=[99,100] -> OpaqueWhereValue-f98ad7d2a26d333e9d25c1bf791b80754ae1acd0

COLLIDE? YES
```

Five `stdClass` objects created and destroyed in sequence produced one distinct hash out of five.

So two different `where()` values in the same request share a cache key. The second query is served the first one's rows. That is the outcome your "Why hash the object rather than name its class" section argues against. The fallback reintroduces it.

Three ways to close it, and I have a preference:

**Option A: make the fallback random per call.** Use `random_bytes()` or `uniqid("", true)` instead of `spl_object_hash()`. One line, and it genuinely delivers what the comment promises. The cost is cache growth. Every such query writes an entry nothing will ever read.

**Option B: throw a clear exception naming the class and the column.** No collision and no cache growth. The developer learns the query cannot be keyed. The cost is a crash, which is what this PR exists to remove. It is a far narrower and louder one.

**Option C: the do-not-cache signal you already offered.** You wrote: *"Not caching such a query at all would be cleaner, but `CacheKey` cannot signal that; it belongs in `Buildable`/`Caching`. Happy to follow up if you want it."* Yes please, and I would rather have it here than as a follow-up.

**I would like Option C.** It is the only one with no bad outcome. Option A trades correctness for unbounded writes, and Option B trades it for a production exception. It also fixes item 3 below at the same time. It is the largest diff of the three, and I think correctness in this class is worth that.

If Option C is more than you want to take on, say so. We can talk about A as a stopgap. I would rather that than have you stuck.

---

## 🟠 2. `hashOpaqueObject()` is not covered by the new test

`testWithInUsingCollectionQueryWithCollectionValues` cannot reach it. The expected key is:

```
...:authors:...author-finances->tags_jsoncontains_["foo","bar"]-authors.deleted_at_null
```

That `["foo","bar"]` is the binding. `getValuesFromBindings()` at `CacheKey.php:738` discards the value segment. It substitutes `stringifyBinding()` whenever the clause owns a binding slot. `whereJsonContains()` binds, so it always does.

`processEnum()` on a `Collection` returns `Illuminate\Support\Collection-<sha1>`. That string appears nowhere in the expected key.

Your own note points at this without quite landing on it: *"The Collection carries the same JSON as the array form and binds the same string, so it must produce the same key."* The key is entirely binding-derived.

So the test proves `processEnum()` stops throwing, which is the reported bug and worth having. It proves nothing about `serialize()`, the fallback, the class-name prefix, or collisions. That matters here because the blocker lives in exactly the function with no coverage.

This repo has a history of cache-key tests that pass without exercising the mechanism. I check for that shape specifically now. Please keep the test you wrote and add one that reaches `hashOpaqueObject()` directly. Two paths put its output into the key with no binding substituting it:

- `stringifyHavingValue()` at `CacheKey.php:290`, reached by `having()`
- `getRowValuesClauses()` at `CacheKey.php:564`, reached by `whereRowValues()`

Either works. Please confirm the new test goes red without the fix, as you did for the first one.

---

## 🟡 3. `serialize()` collapses a resource to `i:0`, so the safe branch can collide too

`serialize()` does not throw on a resource. It writes `i:0` and carries on:

```
A: O:11:"HoldsStream":2:{s:1:"h";i:0;s:2:"id";i:1;}
B: O:11:"HoldsStream":2:{s:1:"h";i:0;s:2:"id";i:1;}
same output for different streams? YES
```

Two objects differing only by a resource handle get one key, without ever reaching the fallback. Your description names `Closure` and PDO as the throwing cases, and resources behave differently from both.

This is much narrower than item 1, and a resource inside a `where()` value is unusual. Option C above resolves it, since such a value would simply not be cached. If you go another route, a comment recording the case is enough.

---

## 🟡 4. The "no array can reach it" note covers three of the five call sites

You wrote:

> `getValuesFromWhere()` flattens first, so no array can reach it today

That holds for the three calls routed through `getValuesFromWhere()`. `processEnum()` has five callers. `stringifyHavingValue()` at `:290` handles arrays itself, so it is fine. `getRowValuesClauses()` at `:564` maps `$where["values"]` straight in.

Laravel's `whereRowValues()` validates only that the column and value counts match, not the element types. So a nested array does reach `processEnum()`, where `"{$value}"` now yields `"Array"` rather than the previous `TypeError`.

Such a query fails at binding time anyway, so this is not dangerous. I would like the sentence corrected, because the next person to touch this function will rely on it.

---

## ✅ What I checked and confirmed correct

I want to be clear about how much of this held up, because most of it did.

| Your claim | Result |
|---|---|
| `Stringable` is auto-implemented for any class declaring `__toString()` | ✅ confirmed |
| `serialize()` throws on a `Closure`-holding object | ✅ `Exception: Serialization of 'Closure' is not allowed` |
| `serialize()` is stable across requests | ✅ identical sha1 across separate PHP processes |
| No existing key changes | ✅ holds, and more strongly than you argued |
| `Expression` still routes to `expressionToString()` | ✅ handled before the new `is_object` branch |
| The new test runs in CI | ✅ every job group declares a `postgres:15-alpine` service |

Two of those are worth expanding on.

`CacheKey.php:3` declares `strict_types=1`. So before this PR, every object outside the union threw, including `Stringable` ones. No key that worked can change. You understated your own safety margin there.

The `Expression` ordering is load-bearing and you got it right. Laravel removed `Expression::__toString()` in version 10, so an `Expression` is not `Stringable` today. Had the explicit branch not preceded your new `is_object` check, every raw-expression key would have changed silently. It does precede it.

---

Thank you again for the care you put into this. The production context you gave made the bug concrete. Once items 1 and 2 are sorted I am happy to take this. Ask me anything if item 1's Option C gets awkward, and I will help work out the shape.

PR #630, opened 2026-09-16
Memoize the morphTo type lookup instead of running it per tag set

bastianschwarz:
`getMorphToTags()` ran a `SELECT DISTINCT` against the parent table on every tag generation, so caching a query cost an uncached query each time.

## The problem

Without a morph map the only way to know which classes a `morphTo` points at is to ask the table:

```sql
select distinct "commentable_type" from "comments" where "commentable_type" is not null
```

Tags are generated once per cached query touching the relation, for reads and for flushes alike. In our application a Filament table listing activity-log rows fired it three or four times per request.

## The change

The result is memoized through `ModelCacheRepository::rememberForever()`, tagged with the table it reads via `makeTableTags()`, so the parent model's own prefix and connection are respected.

A new morph type can only arrive through a write to that table, and that write invalidates the table tag, so the memo is dropped exactly when it stops being true. A static would hold the snapshot for the life of the process and leak it between requests, queue jobs and tests.

No `TaggableStore` guard: `ModelCacheRepository` already handles that. An untaggable store gets the entry untagged, and `invalidateTags()` flushes the whole repository there, so the memo goes with it.

## Testing

New `MorphTypeLookupTest`, existing fixtures only:

- `testMorphTypeLookupIsNotRepeatedForEveryTagGeneration` — three cached queries, one lookup. Red before at `Failed asserting that 3 is identical to 1`.
- `testMorphTypeLookupIsRepeatedAfterTheTableIsWrittenTo` — a warm memo answers without a query, then a comment of a new type is written and the lookup runs again. Without this the first test would also pass for a memo that never expires.

Reverted the change, confirmed both go red, restored it.

After: `Tests: 545, Assertions: 1335, Skipped: 4`

`Cache::memo()` would collapse the remaining reads within a request, but `MemoizedStore` is not a `TaggableStore`, so the entry would lose its tag and outlive the write that should drop it. Happy to look at an in-request layer on top of the tagged entry separately.


mikebronner (changes requested):
👋 Thank you for this. The problem is real, the diagnosis is right, and the tests are the strongest part.

I pulled your test file and ran it against master at `53f724a`. Both tests failed for the stated reason:

```
1) testMorphTypeLookupIsNotRepeatedForEveryTagGeneration
   Failed asserting that 3 is identical to 1.
2) testMorphTypeLookupIsRepeatedAfterTheTableIsWrittenTo
   Failed asserting that 1 is identical to 0.
```

The first message matches your description word for word. The second test is the one I want to call out. Your note explains why it exists: *"Without this the first test would also pass for a memo that never expires."* That is exactly right, and it is the trap this repo has walked into before. Writing the second test unprompted is the reason I trust the first one.

Two changes requested, plus one note that needs nothing from you.

---

## 🟠 1. `rememberForever` turns a self-healing lookup into permanent staleness

Your invalidation argument reads:

> A new morph type can only arrive through a write to that table, and that write invalidates the table tag, so the memo is dropped exactly when it stops being true.

The first half is true. The second half holds only for writes through a **cachable model**. These four do not flush the table tag:

- `DB::table('comments')->insert(...)`
- a bulk import or a seeder
- a model without the `Cachable` trait
- another service sharing the database

Before this change the `SELECT DISTINCT` ran on every tag generation. A new morph type was picked up on the very next query. With `rememberForever` the wrong answer is held with no expiry.

The cost is not a slow query. A missing morph type means cached queries omit that type's table tag. Writes to that model then never invalidate those queries. Stale rows are served until someone flushes the cache by hand.

So the change trades a per-call query for a failure mode that no longer heals itself. That trade is worth making. I would just like it bounded.

**Please use a TTL rather than `rememberForever`.** Sixty seconds removes essentially all of the cost you are targeting. It also caps the damage at sixty seconds instead of forever. If you would rather pick a different window, say what and why, and I am happy to take it.

---

## 🟡 2. The memo key and its tag are built from different models

```php
ModelCacheRepository::make()->rememberForever(
    $this->getCachePrefix() . "morph-types:" . $table . ":" . $column,
    $this->makeTableTags([["table" => $table, "model" => $relation->getParent()]]),
```

`getCachePrefix()` reads `$this->model`, the querying model. `makeTableTags()` reads `$relation->getParent()`, the model owning the morph column. Those are the same object only when the relation is queried directly.

`getMorphToTagsForRelation()` walks `explode('.', $relationName)` starting from `$this->model`. So `Post::with('comments.commentable')` gives `$this->model = Post` and `$relation->getParent() = Comment`. The key then carries Post's prefix while the tag carries Comment's.

This is not a staleness bug. The tag is the correct one, so invalidation still works. It fragments the memo instead: one entry per querying model rather than one per table. Your stated benefit shrinks by however many models reach the relation.

Your tests cannot see this, because they query `Comment` directly and the two models coincide there.

**Please key off `getCachePrefixForModel($relation->getParent())` instead**, so the key and the tag name the same model. A test through a nested eager load would pin it.

---

## ℹ️ 3. A note that needs nothing from you

`getMorphToTags()` tags its results with `$this->getCachePrefix()`, the querying model's prefix. `CachePrefixing`'s own docblock describes that as the wrong prefix for a related table:

> Tagging a related table with the *querying* model's prefix therefore produces a tag the related model's own write never flushes, and the cached query stays stale for as long as the entry lives.

That predates your PR and is mine to fix. I mention it only so the inconsistency around your new code does not read as something you introduced.

---

## 🔧 A rebase is needed, and that one is on me

I merged a cleanup of `src/CacheTags.php` after you opened this. The PR now reports `CONFLICTING`. Sorry about the timing.

The conflict should be mechanical. The exact block you replace is byte-identical on master:

```php
$types = $relation->getParent()
    ->newQuery()
    ->getQuery()
    ->select($column)
    ->from($table)
    ->whereNotNull($column)
    ->distinct()
    ->pluck($column)
    ->toArray();
```

Only the surrounding context moved. Of the five PRs you have open, this is the only one affected.

CI has not run on this branch yet. The rebase will trigger it.

---

## ✅ What I checked and confirmed correct

| Your claim or design choice | Result |
|---|---|
| Both tests go red without the change | ✅ reproduced on master at `53f724a` |
| The memoized query does not recurse into the cache | ✅ the callback drops to `->getQuery()` |
| A cache-backed memo beats a static | ✅ agreed, and the reasoning is right |
| `ModelCacheRepository` already handles an untaggable store | ✅ confirmed |
| No cross-connection key collision | ✅ `use-database-keying` defaults to `true` |

Your reasoning about `Cache::memo()` is also right. `MemoizedStore` is not a `TaggableStore`, so the entry would lose its tag and outlive the write. Leaving it out was the correct call. I would be glad to look at an in-request layer on top of the tagged entry separately.

---

Thank you again for finding this. The Filament detail in your description told me immediately how it shows up in practice. Once the TTL and the key prefix are sorted I am happy to take this. Ask me anything if either one turns out more awkward than it looks.

PR #631, opened 2026-09-16
Tag the tables a where-subquery joins, not only the one it selects from

bastianschwarz:
A `whereHas()` over a `belongsToMany` was never tagged with the pivot table its subquery joins, so a write to that pivot left the query serving stale rows.

## The gap

`#623` and `#626` tag the table a where-subquery selects `from`, and `getJoinTags()` tags the outer query's joins. Nothing tags a join sitting inside a subquery:

```php
foreach ($builder->joins ?? [] as $join) {
    foreach ($join->wheres ?? [] as $where) {
        $tables = array_merge($tables, $this->getSubqueryTablesFromWhere($where, $seen));
    }
}
```

That walks each join's condition for further nested subqueries, but never records `$join->table`. A `belongsToMany` subquery selects `from` the related table and joins the pivot to connect it to the outer one, and the pivot decides which rows come back exactly as the related table does.

## How it showed up

A Filament select listing the records not yet assigned to the owner — `whereDoesntHaveRelation('whitelabels', 'whitelabel_id', $owner)`. After assigning one, the select still offered it, because the write to the join table could not bust the cached options query. The form's own `in` rule then rejected the submission, so the user was offered a choice the form refused.

## The change

Record `$join->table` alongside walking its wheres, alias-stripped. A join whose table is an `Expression` (`joinSub()` and friends) names no table and would raise a `TypeError` in `stripTableAlias()`, so it is skipped here exactly as `getJoinTags()` skips it.

## Testing

`SubqueryTagTest::testSubqueryJoinTagsTheJoinedTable` — `(new Book)->whereHas("stores")` must be cached under the book tags plus `stores` and `book-store`.

That file's `assertCachedUnderTags()` asserts the complete tag set, since `TagSet` namespaces an entry by the tags it was written with, so the test proves `book-store` was the only tag missing. Red before with `Failed asserting that null is not null`.

`CachedBuilderRelationshipsTest::testHasRelationshipResults` hardcodes the tag list for `(new Book)->with("stores")->has("stores")` and gains the pivot tag — the same one-line update `#623` and `#626` made to their neighbours.

After: `Tests: 544, Assertions: 1333, Skipped: 4`

Verified against our own application as well: the two admin tests that reproduced this now pass with model caching enabled.


mikebronner (changes requested):
👋 Thank you for this. The gap is real, and I reproduced everything you claimed about it.

I also ran CI, which had not fired on this branch before. All thirteen checks pass, and the PR is now `MERGEABLE / CLEAN`. So your suite numbers are confirmed independently.

Most of this review is confirmation. There is one change I want before it lands, and one question I need your help with.

---

## ✅ What I checked and confirmed

| Your claim | Result |
|---|---|
| Nothing tags a join sitting inside a subquery | ✅ confirmed on master, the loop walks `$join->wheres` and never records `$join->table` |
| The new test is red without the change | ✅ 1 failure of 22, `Failed asserting that null is not null`, word for word |
| The relationships test is red without the change | ✅ errors on the missing tag |
| The `Expression` guard matches `getJoinTags()` | ✅ same `is_string()` check, same reason |

I also checked every test file carrying a hardcoded tag list, twenty of them, for relation-existence queries. Only a `belongsToMany` subquery carries a join. `testWhereHasRelationship` asserts results rather than tags. **Your single test update is complete.** That is easy to get wrong and you got it right.

One worry of mine turned out to be nothing, and it is worth recording so nobody re-raises it. `make()` merges `$joinTags` before `$subqueryWhereTags` and then calls `->unique()`. An outer join that was already tagged keeps its position and the duplicate is dropped. So a query with joins but no subquery joins sees no tag change at all.

---

## 🟠 1. Please add an invalidation test

This is the one change I want before merging.

Your opening line is:

> a write to that pivot left the query serving stale rows

`testSubqueryJoinTagsTheJoinedTable` asserts that `book-store` is in the tag set. It does not write to the pivot and then check the query returns fresh rows. So it proves the tag is present. It does not prove the tag does anything.

The repo already has this pattern in `BelongsToManyTest::testInvalidatingCacheWhenDetaching` and in `CustomPivotCacheInvalidationTest`. A test in that shape would demonstrate the behaviour the PR exists to deliver. It would also answer item 2 below.

---

## 🟡 2. Which write is the new tag protecting against?

This is a genuine question. You have the failing application and I do not, so I would rather ask than guess.

Laravel invalidates a tagged entry when **any** tag in its set is flushed. So a new tag only adds protection when some write flushes that tag. It also has to flush nothing else in the set.

Reading the flush paths, `attach()` and `detach()` land here:

```php
protected function flushCacheForPivotOperation(): void
{
    if (method_exists($this->parent, 'flushCache')) {
        $this->parent->flushCache();
    }
    $relatedModel = $this->getRelated();
    if (method_exists($relatedModel, 'flushCache')) {
        $relatedModel->flushCache();
    }
}
```

That flushes the parent and the related model. `flushCache()` builds its tags from an empty query. Each one therefore flushes its own class tag and table tag. Both of those were already in the cached query's tag set before this PR.

So on that path the query looks like it was already being busted. The new tag reads as load-bearing only when a `Cachable` pivot model is saved directly. `PivotModelObserverTest` already covers that case.

So my question is this.

What was the write path in your application?

If it is something other than `attach()` or `detach()`, that is the test worth writing. It would prove the fix outright.

The tag may turn out to be correct but not load-bearing. I am still happy to take the change in that case. A pivot genuinely decides which rows come back, so tagging it is right on principle. I would just want the description to say that, rather than claim a fix the tests cannot show.

---

## 🟡 3. Worth a line in the comment: the pivot tag takes the querying model's prefix

The new entry is recorded with `"model" => null`, so `makeTableTags()` falls back to `$this->model`. `CachePrefixing`'s own docblock spells out what that costs:

> Tagging a related table with the *querying* model's prefix therefore produces a tag the related model's own write never flushes, and the cached query stays stale for as long as the entry lives.

So a pivot on another connection is still tagged under the wrong prefix. The same goes for one whose model declares its own `$cachePrefix`. That tag is never flushed by the pivot's own write.

This matches how `getJoinTags()` already behaves, so it is consistent rather than wrong. Please note it in the comment. The next reader should not take the class to be closed.

---

## ✅ 4. The cache-tag churn is expected, and needs nothing from you

Adding a tag changes the tag set, so every `whereHas` over a `belongsToMany` reads cold after upgrading. I am recording it here for completeness rather than asking you to act on it. That churn is the normal price of a tag correctness fix here. I am fine with it.

---

## ℹ️ 5. Two small accuracy items

The comment says these are "the ones nested inside a subquery". The branch also runs on the outer query's joins, because `getSubqueryWhereTags()` passes `resolveBaseQuery()`. That is harmless for the reason given above. The comment just describes a narrower reach than the code has.

The shape you hit in production is `whereDoesntHaveRelation`, a `NotExists`. The new test covers `whereHas`, an `Exists`. Both route through the same branch, so the risk is low. The shape you actually reported is still untested. If the invalidation test in item 1 uses `whereDoesntHaveRelation`, it closes both at once.

---

Thank you again for tracking this down. The Filament detail was the useful part of the description. A user offered a choice that the form then refuses is a very specific symptom. It made the gap easy to place. Answer item 2 and add the test from item 1, and I am happy to take this.

mikebronner:
📌 Correcting my review. Item 2 was answered before I asked it. Item 1 was asking for something that cannot pass yet.

You filed the answer yourself in #634, four days before I raised the question. I should have read the open issues first. Apologies for the round trip.

---

## ✅ Item 2 is withdrawn

I asked which write path the new pivot tag protects against. The `attach()` and `detach()` path looked like it was already busting the query.

The answer is in #634, and it is better than my guess. The masking comes from the `saved` event listener that `bootCachable()` registers. It does not come from the pivot path. `withoutEvents()`, `saveQuietly()` and `createQuietly()` switch that listener off. On those writes nothing invalidates the read.

So the question was already answered. Please ignore it.

---

## 🔄 Item 1 changes from a request to a note

I asked for a test that writes to the pivot and shows the query returns fresh rows. That test cannot pass today, and not because of anything in this PR.

I checked all three flush paths on master:

- `flushCacheForPivotOperation()` flushes the parent and the related model.
- `flushCache()` builds its tags from an empty query. It flushes a class tag and a table tag.
- `flushRelationshipCache()` flushes the related model's tags, built the same way.

None of them flushes the pivot table's tag. That is exactly the gap #634 describes.

So your new tag is correct and currently inert. It is the read half. #634 is the write half. Neither alone busts the query. That is also why you could not build a failing reproduction for #634.

**I am not holding this PR for that test.**

Adding the tag now is right. It is what makes the #634 fix work when it lands.

---

## 🧭 How the two fit together, for the record

They are different query shapes and both gaps are real.

A plain relation read joins the pivot in the outer query, so `getJoinTags()` already tags it. That is the shape #634 is about, and its gap is purely on the flush side.

A `whereHas()` puts that join inside a subquery, which nothing walked. That is the shape this PR is about. Its gap is on the read side.

---

## 📋 What still stands, and both are one-liners

**Item 3.** The new entry is recorded with `"model" => null`, so the pivot tag takes the querying model's prefix. A pivot on another connection is still tagged under the wrong prefix. So is one whose model declares its own `$cachePrefix`. That prefix is never flushed by the pivot's own write. This matches `getJoinTags()`, so please just note it in the comment.

**Item 5.** The comment says these are the joins nested inside a subquery. The branch also runs on the outer query's joins, because `getSubqueryWhereTags()` passes `resolveBaseQuery()`. That is harmless, and the comment describes a narrower reach than the code has.

The `NotExists` point from item 5 also softens. I suggested the invalidation test use `whereDoesntHaveRelation`. That test is no longer requested. A second tag-assertion test in that shape would do, and I am happy without one.

Item 4 needed nothing from you and still does not.

Sort those two comment lines and I will approve.

---

Thank you for the patience, and for filing #634 with that much measurement in it. You logged one `select` for the first `assignRole()` and none for the second. That is what made the masking mechanism legible. I will reply on #634 separately about which direction I want there.

PR #633, opened 2026-09-16
Build a cache key for a query whose from is a subquery

bastianschwarz:
`getTableSlug()` passed `$this->query->from` straight to `Str::slug()`, which casts its argument to string. `fromSub()` and `fromRaw()` put an `Expression` there, so building the key raised an `Error`.

## The problem

```
Error: Object of class Illuminate\Database\Query\Expression could not be converted to string
src/CacheKey.php:646
```

`Builder::from()` routes anything queryable to `fromSub()`, so `from` holds either a table name or an `Expression`. There is no `fromSub` or `fromRaw` anywhere in `tests/` today, so the second case was uncovered.

## A second bug behind the first

Fixing the cast alone is not enough. A subquery compiles to SQL holding placeholders:

```sql
(select * from "authors" where "id" > ?) as "authors"
```

so `fromSub(id > 1)` and `fromSub(id > 2)` produce byte-identical SQL and differ only in their bindings. Those bindings reach the key nowhere else: `bindings` is in `UNKEYED_INFRASTRUCTURE_PROPERTIES`, and `getBindingsSlug()` reads a freshly built model query rather than this one.

Hashing the statement alone therefore gives two different queries one key and serves one of them the other's rows. `testTwoDifferentFromSubQueriesDoNotShareAKey` failed on exactly that once the cast was fixed.

## The change

A derived table is keyed by `"sub-" . sha1($sql . $fromBindings)`, so the key stays a fixed length whatever the statement's size. A plain table name keeps the slug it has always had, so no existing key changes.

## Testing

New `FromSubTest`:

- `testFromSubQueryIsCached` — the query runs and matches its uncached twin. Red before with the `Error` above.
- `testTwoDifferentFromSubQueriesDoNotShareAKey` — two subqueries differing only in a bound value get different keys, and the second still returns its own rows after the first has been cached.

After: `Tests: 545, Assertions: 1336, Skipped: 4`


mikebronner (changes requested):
👋 Thank you for this. The fix is correct, and the second bug you found behind the first is the better catch.

I am requesting one change. It is about what the fix newly makes reachable, not about the fix itself.

---

## ✅ What I checked and confirmed

Both new tests error on clean master at exactly `CacheKey.php:646`:

```
Error: Object of class Illuminate\Database\Query\Expression could not be converted to string
```

Same message and line you quote. CI is 13 of 13.

| Your claim | Result |
|---|---|
| `bindings` sits in `UNKEYED_INFRASTRUCTURE_PROPERTIES` | ✅ line 74 |
| The `from` bindings reach the key nowhere else | ✅ existing `getChannelBindingsSlug()` calls cover `having`, `groupBy` and `join` only |
| A plain table name keeps its slug | ✅ the new branch is gated on `instanceof Expression` |

Your second test is a real discriminator. Both queries compile to byte-identical SQL. They differ only in a bound value. So different keys can only come from the bindings entering the key.

Two details I want to name. The test calls `ReflectionMethod::invoke()` without `setAccessible()`. That avoids a PHP 8.5 deprecation the rest of our suite still triggers. It also carries `declare(strict_types=1)`, matching the convention I introduced in `53f724a`.

---

## 🟠 Please also record the table the subquery reads

`CacheTags::make()` never reads `from`. I traced every contributor to the tag set:

- `getTagName()` and `getTableTagName()` read `$this->model`, not the query.
- `getJoinTags()` reads `$this->query->joins`.
- `getSubqueryWhereTags()` walks `wheres`, `joins` and `unions`.

`CachedQueryBuilder` overrides `whereIn`, `joinSub`, `leftJoinSub`, `rightJoinSub`, `joinLateral` and `leftJoinLateral`. It does not override `fromSub` or `from`. So the table is never recorded there either.

That leaves this query tagged `authors` while it reads `books`:

```php
(new Author)->fromSub(fn ($q) => $q->from("books")->where(...), "b")->get();
```

A write to `Book` never busts it, and the next read is stale.

Why this belongs on this PR rather than in a follow-up. Before your change, `fromSub()` on a cachable model raised an `Error`, so nobody could reach the stale path. After it, the query caches. The change turns a loud crash into a silent stale read, for exactly the case it enables.

I would rather not ship a release where that is true.

---

## 🔧 The pieces are already in the file

`CachedQueryBuilder`'s own docblock describes this problem for `joinSub()`:

> `joinSub()` compiles its subquery to SQL and hands `join()` an `Illuminate\Database\Query\Expression`, so the `JoinClause` names no table and `CacheTags` has nothing to tag

`recordRelatedWhereSubqueryTable()` already does the recording. `recordSubqueryTable()` already handles a Closure, a Relation, an Eloquent builder and a query builder. An override of `fromSub()` routing `$query` through it is a few lines.

`SubqueryTagTest::testWhereInSubqueryTagsTheSubqueryTable` is the test template, and `assertCachedUnderTags()` asserts the complete tag set.

I traced the chain end to end and it closes. `resolveBaseQuery()` returns the `CachedQueryBuilder`. `getSubqueryTablesFromBuilder()` reads `getRelatedSubqueryTables()` from it. `makeTableTags()` then emits the tag that `Book::save()` flushes.

---

## 📐 What this will and will not cover

Worth stating so the comment can say it too.

It covers the common case. That is a source model sharing a connection with the querying model. It must also declare no `$cachePrefix`.

It does not cover a source model on another connection, or one with its own `$cachePrefix`. The table is recorded with no model, so the tag keeps the querying model's prefix. `recordRelatedWhereSubqueryTable()`'s docblock already states this limitation for `whereIn()`, and it applies here for the same reason.

It cannot cover `fromRaw()`. The table name is inside a raw string and is never addressable. That matches how `getJoinTags()` gives up on an `Expression` join table.

Please note both limits in the comment, so the next reader does not take the class to be closed.

---

Thank you again. Finding the binding collision while fixing the cast is the part that impressed me. Fixing the cast alone would have shipped a key that silently served one query another's rows. That is the worst outcome this package has. Add the `fromSub()` recording with a tag test and I am happy to take this.

